PR #8, opened 2022-03-17
Source config struct

lovromazgon:
### Description

This PR introduces a source config struct to make configuration handling a bit nicer.

### Quick checks:

- [X] I have followed the [Code Guidelines](https://github.com/ConduitIO/conduit/blob/main/docs/code_guidelines.md).
- [X] There is no other [pull request](https://github.com/ConduitIO/conduit-connector-postgres/pulls) for the same update/change.
- [X] I have written unit tests.
- [X] I have made sure that the PR is of reasonable size and can be easily reviewed.

PR #9, opened 2022-03-17
Tests refactoring

lovromazgon:
### Description

This PR refactors tests and makes them a bit cleaner, we now create a table with a random name at the start, use it for the duration of the test and drop it after the test. Assertions are now made with `matryer/is` instead of `google/go-cmp`.

Based on #8.

### Quick checks:


dylanlott (approved):
LGTM

PR #16, opened 2022-03-24
adds append only handling when no table key is set

dylanlott:
### Description

Attempts to append-only insert records if no primary key is set for the table. If one is set, it will upsert as usual.

### Quick checks:

### DIFF
--- a/destination/destination_test.go
+++ b/destination/destination_test.go
@@ -54,7 +54,7 @@ func TestAdapter_Write(t *testing.T) {
 					Position: sdk.Position("5"),
 					Metadata: map[string]string{
 						"action": "insert",
-						"table":  "records1",
+						"table":  "keyed",
 					},
 					Key: sdk.StructuredData{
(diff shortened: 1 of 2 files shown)


destination/destination_test.go
```diff
@@ -94,6 +94,54 @@ func TestAdapter_Write(t *testing.T) {
...
+			},
+			args: args{
+				ctx: context.Background(),
+				record: sdk.Record{
+					Position: sdk.Position("28357"),
+					Metadata: map[string]string{
+						"table": "unkeyed",
+					},
+					Key: sdk.StructuredData{
+						"key": "appendonly",
```
hariso, line 113: What's `key` here? Is it the column name?

dylanlott: `key` in this case is the name of the column that corresponds to the key, but the table has no key in this case. This is to test that if a key is provided in a record, it is still recorded, even if the table has no primary key set.

Issue #15, opened 2022-03-24
Snapshot support in logrepl mode


lovromazgon:
When using logical replication mode it should be possible to create a snapshot first.

PR #22, opened 2022-03-30
adds snapshot iterator

dylanlott:
### Description

Adds the snapshot iterator to the `logrepl` package. 

This is the first step towards adding a snapshot that works seamlessly with CDC mode.

We define a `SnapshotIterator` that takes a config which allows for a `SnapshotName` to be passed and iterates over
the rows from the configured table at that snapshot's point in time.

We necessarily secure a transaction with `READ REPEATABLE` from the database for our iterator to read.

Related to #15 

### Quick checks:


source/logrepl/snapshot.go
```diff
@@ -0,0 +1,221 @@
...
+	err = s.loadRows(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load rows: %w", err)
+	}
+
+	return s, nil
+}
+
+func (s *SnapshotIterator) attachPool(ctx context.Context) error {
+	pool, err := pgxpool.Connect(ctx, s.config.URI)
```
lovromazgon, line 79: The iterator shouldn't open a new connection, it should get a connection in the constructor and use it.

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		return err
+	}
+	s.pool = pool
+	return nil
+}
+
+func (s *SnapshotIterator) loadRows(ctx context.Context) error {
+	snapshotTx := fmt.Sprintf("SET TRANSACTION SNAPSHOT '%s';", s.config.SnapshotName)
+	_, err := s.tx.Query(ctx, snapshotTx)
```
lovromazgon, line 89: We should use `Exec` instead of `Query`, then we don't have to worry about closing the rows (that's what's missing right now and is causing the conn busy error).

source/logrepl/snapshot_test.go
```diff
@@ -0,0 +1,155 @@
...
+		t.Log(rec)
+		rec.CreatedAt.After(now)
+	}
+
+	i.NoErr(s.Teardown(ctx))
+}
+
+func createTestSnapshot(t *testing.T, db *pgx.Conn) string {
+	var n *string
+	query := `SELECT * FROM pg_catalog.pg_export_snapshot();`
```
lovromazgon, line 147: According to [Postgres docs](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-SNAPSHOT-SYNCHRONIZATION):
> Saves the transaction's current snapshot and returns a text string identifying the snapshot. This string must be passed (outside the database) to clients that want to import the snapshot. **The snapshot is available for import only until the end of the transaction that exported it.**

So to set up the test correctly we need to 1. open a separate connection, 2. open a transaction in that connection and 3. export a snapshot. The transaction and connection should stay open at least until we create the snapshot iterator.

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,207 @@
...
+	return nil
+}
+
+func (s *SnapshotIterator) Next(ctx context.Context) (sdk.Record, error) {
+	if !s.rows.Next() {
+		if err := s.rows.Err(); err != nil {
+			return sdk.Record{}, fmt.Errorf("rows error: %w", err)
+		}
+
+		return sdk.Record{}, sdk.ErrBackoffRetry
```
lovromazgon, line 111: `ErrBackoffRetry` doesn't seem like the right error here, if there are no more records there's no reason to retry. We should probably create an error to signal that the snapshot reached the end, this way the combined iterator can check for the error and switch to the CDC iterator.

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,207 @@
...
+	return nil // noop for snapshots
+}
+
+// Teardown attempts to gracefully teardown the iterator.
+func (s *SnapshotIterator) Teardown(ctx context.Context) error {
+	s.rows.Close()
+	return s.commit(ctx)
+}
+
+func (s *SnapshotIterator) commit(ctx context.Context) error {
```
lovromazgon, line 128: If we switch to `pgx.Tx` we can remove this function and use `pgx.Tx.Commit` instead.

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	r.CreatedAt = time.Now()
+
+	r.Metadata = map[string]string{
+		"action": actionSnapshot,
+		"table":  s.config.Table,
+	}
+
+	r.Position = sdk.Position(strconv.FormatInt(s.internalPos, 10))
```
lovromazgon, line 153: I'm thinking how we could make positions a bit more unique. We don't want to hit [this error](https://github.com/ConduitIO/conduit/blob/77abccad616f11b89e154292c7ea37080d17f4bb/pkg/pipeline/stream/acker.go#L163) when we run two postgres source connectors in the same pipeline. We could at least prefix the position with the table name.

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Ack is a noop for snapshots
+func (s *SnapshotIterator) Ack(ctx context.Context, pos sdk.Position) error {
+	return nil // noop for snapshots
+}
+
+// Teardown attempts to gracefully teardown the iterator.
+func (s *SnapshotIterator) Teardown(ctx context.Context) error {
+	s.rows.Close()
```
lovromazgon, line 124: After this we should probably check `s.rows.Err()`.

source/logrepl/snapshot_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"time"
+
+	"github.com/conduitio/conduit-connector-postgres/test"
+	sdk "github.com/conduitio/conduit-connector-sdk"
+
+	"github.com/matryer/is"
+)
+
+func TestLifecycle(t *testing.T) {
+	i := is.New(t)
```
lovromazgon, line 29: Nitpick: by convention we use `is := is.New(t)`, I know we are overwriting the package name `is` but after this call we actually don't have a use for the package anymore.

source/logrepl/snapshot_test.go
```diff
@@ -0,0 +1,87 @@
...
+			"column2": int32(123),
+			"column3": bool(false),
+		},
+		Metadata: map[string]string{
+			"action": actionSnapshot,
+			"table":  table,
+		},
+	})
+
+	err = s.Teardown(ctx)
```
lovromazgon, line 82: Here I would actually expect to get an error, we didn't range over all the rows in the snapshot, this means that the next time we start the pipeline we won't be able to start at the same place we left off.

Also, would be nice to see a test where we actually range over the whole snapshot.

lovromazgon (changes requested):

source/logrepl/snapshot_test.go
dylanlott: addressed in #27

source/logrepl/snapshot.go
dylanlott: I decided to use table names for now, if we need to increase specificity in the future we can add the database name to the string fairly easily.

source/logrepl/snapshot.go
dylanlott: Addressed in #27.

source/logrepl/snapshot_test.go
```diff
@@ -0,0 +1,113 @@
...
+	tx, err := conn.Begin(ctx)
+	is.NoErr(err)
+	query := `SELECT * FROM pg_catalog.pg_export_snapshot();`
+	rows, err := tx.Query(context.Background(), query)
+	is.NoErr(err)
+
+	var name *string
+	is.True(rows.Next())
+	err = rows.Scan(&name)
+	is.NoErr(err)
```
lovromazgon, line 94: Nitpick: we can simplify using `QueryRow`, we know there will be only one row. Also `name` doesn't have to be a pointer, and we can use `ctx` as the context.
```suggestion
	row := tx.QueryRow(ctx, query)
	var name string
	err = row.Scan(&name)
	is.NoErr(err)
```

source/logrepl/snapshot_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/matryer/is"
+)
+
+func TestLifecycle(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+
+	pool := test.ConnectPool(ctx, t, test.RegularConnString)
+	table := createTestTable(t, pool)
```
lovromazgon, line 35: This can be simplified further, `test.SetupTestTable` already does everything we need:
```suggestion
	table := test.SetupTestTable(ctx, t, pool)
```

source/logrepl/snapshot_test.go
```diff
@@ -0,0 +1,113 @@
...
+	conn, err := pool.Acquire(ctx)
+	is.NoErr(err)
+	s, err := NewSnapshotIterator(context.Background(), conn.Conn(), SnapshotConfig{
+		SnapshotName: name,
+		Table:        table,
+		Columns:      []string{"id", "key", "column1", "column2", "column3"},
+		KeyColumn:    "key",
+	})
+	is.NoErr(err)
+	t.Cleanup(func() { conn.Release() })
```
lovromazgon, line 47: Nit: simplification
```suggestion
	t.Cleanup(conn.Release)
```

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,213 @@
...
+		if rollErr := s.tx.Rollback(ctx); rollErr != nil {
+			sdk.Logger(ctx).Err(rollErr).Msg("set transaction rollback failed")
+		}
+		return fmt.Errorf("failed to set transaction snapshot id: %w", err)
+	}
+
+	return nil
+}
+
+func (s *SnapshotIterator) Next(ctx context.Context) (sdk.Record, error) {
```
lovromazgon, line 113: One more important thing I forgot last time - if the connector receives a stop signal this context will get cancelled but it will continue calling the`Read` method to give the connector a chance to empty its cache. Long story short - we need to check if the context is cancelled and return `ctx.Err()` if it is, to ensure the connector can stop in the middle of a snapshot.

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	err := s.startSnapshotTx(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to start snapshot tx: %w", err)
+	}
+
+	err = s.loadRows(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load rows: %w", err)
```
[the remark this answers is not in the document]
lovromazgon: This issue still exists, if we fail to load rows there is an open transaction in the DB.

lovromazgon (changes requested):
A couple of nitpicks, but the main things are the open transaction and the unchecked context in `Next`, after this we should be good 👍  The error related issues will be addressed in #27 so I skipped those.

source/logrepl/snapshot_test.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+	})
+	is.NoErr(s.Teardown(ctx)) // TODO: Should return an error
+}
+
+// createTestSnapshot starts a transaction that stays open while a snapshot run.
+// Otherwise, postgres deletes the snapshot as soon as this tx commits,
+// an our snapshot iterator won't find a snapshot at the specifiedname.
+// https://www.postgresql.org/docs/current/sql-set-transaction.html
+func createTestSnapshot(t *testing.T, pool *pgxpool.Pool) string {
```
lovromazgon, line 80: Nice helper function 👍  The godoc has a couple of typos though 😉

source/logrepl/snapshot_test.go
dylanlott: We actually can't use `QueryRow` here, because it messes with our transaction when we call `Scan`, which handles its own connection cleanup.

lovromazgon: I tried it out and it works for me 🤔

dylanlott: So, I tested this before I wrote my first comment to verify that behavior still persisted, and it broke exactly as I expected it to the first time... but after you said you tested it and it worked for you, I went back and tested it again, and... it worked? 🤔 I guess I'll accept a passing test 😅

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	err := s.startSnapshotTx(ctx, conn)
+	if err != nil {
+		return nil, fmt.Errorf("failed to start snapshot tx: %w", err)
+	}
+
+	err = s.loadRows(ctx)
+	if err != nil {
+		if rollErr := s.tx.Rollback(ctx); err != nil {
```
lovromazgon, line 64:
```suggestion
		if rollErr := s.tx.Rollback(ctx); rollErr != nil {
```

source/logrepl/snapshot.go
```diff
@@ -0,0 +1,219 @@
...
+			sdk.Logger(ctx).Err(err).Msg("load rows failed")
+			return nil, fmt.Errorf("rollback failed: %w", rollErr)
```
lovromazgon, line 66: Nitpick: I'd rather turn the logic around and just log the rollback error (same as we do in line 107)
```suggestion
			sdk.Logger(ctx).Err(rollErr).Msg("rollback failed")
```

lovromazgon (approved):
LGTM 👍  Left two more comments, but approving already!

PR #25, opened 2022-04-05
Update readme and specs

lovromazgon:
### Description

This PR makes sure specs and readme contain the correct information about the config parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,143 +1,110 @@
-# Conduit PostgreSQL Connectors 
+# Conduit Connector PostgreSQL
 
 # Source
-The Postgres Source Connector connects to a database with the provided `url` and
-then will call `Ping` to test the connection. If the `Ping` fails, the `Open`
-method will fail and the Connector won't be started.
 
-Upon starting, the source takes a snapshot of a given table in the database, 
-then switches into CDC mode. In CDC mode, the plugin reads from a buffer of 
-CDC events.
+The Postgres Source Connector connects to a database with the provided `url` and starts creating records for each change
+detected in a table.
+
+Upon starting, the source takes a snapshot of a given table in the database, then switches into CDC mode. In CDC mode,
+the plugin reads from a buffer of CDC events.
 
 ## Snapshot Capture
-When the connector first starts, snapshot mode is enabled. The connector
-acquires a read-only lock on the table, and then reads all rows of the table 
-into Conduit. Once all of the rows in that initial snapshot are read, then the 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,143 +1,110 @@
-# Conduit PostgreSQL Connectors 
+# Conduit Connector PostgreSQL
```
hariso, line 1: Nitpick: I just noticed this, and to me it's more natural to say: a Pg connector, a Kafka connector etc. Or: Conduit connector for Pg.

README.md
```diff
@@ -1,143 +1,110 @@
...
-connector releases its lock and switches into CDC mode. 
 
-This behavior is enabled by default, but can be turned off by adding 
-`"snapshot":"off"` to the Source configuration.
+When the connector first starts, snapshot mode is enabled. The connector acquires a read-only lock on the table, and
+then reads all rows of the table into Conduit. Once all rows in that initial snapshot are read the connector releases
+its lock and switches into CDC mode.
+
+This behavior is enabled by default, but can be turned off by adding `"snapshotMode":"never"` to the Source
+configuration.
```
hariso, line 18: Nitpick: it appears that we're using different line lengths in our IDEs. It would be nice to agree on one and use it (otherwise we might see noise in future PRs).

README.md
```diff
@@ -1,143 +1,110 @@
...
 ```
 
-### CDC Event Buffer
-There is a private variable bufferSize that dictates the size of the channel 
-buffer that holds WAL events. If it's full, pushing to that channel will be a 
-blocking operation, and thus execution will stop if the handler for WAL events 
-cannot push into that buffer. That blocking execution could have unknown 
-negative performance consequences, so we should have this be sufficiently high 
-and possibly configured by environment variable.
-
```
hariso, line 64: I find this information useful, especially since it highlights a possible performance bottleneck. It would be good to keep it somewhere.

hariso (approved):
LGTM. :+1: I think that some of the parts which were removed are worth keeping (somewhere, if the ReadMe is not the best place).

README.md
lovromazgon: Yeah I noticed this as well, I used this just because we use the same format in other connectors as well and to keep it consistent. I'm all for changing this title, but then we should do it across the board. I'd say we can do this after the release since it's only the readme.

README.md
lovromazgon: I removed it because this is outdated information, there is no such variable.

PR #27, opened 2022-04-06
returns snapshot incomplete error when snapshot is interrupted

dylanlott:
### Description

Returns an `ErrSnapshotInterrupt` when a snapshot is stopped before it's finished.
Creates the `ErrSnapshotComplete` variable in anticipation of creating the `CombinedIterator`

Addresses https://github.com/ConduitIO/conduit-connector-postgres/pull/22#discussion_r843033129 

### Quick checks:


source/logrepl/snapshot.go
```diff
@@ -133,7 +141,15 @@ func (s *SnapshotIterator) Teardown(ctx context.Context) error {
 	if commitErr := s.tx.Commit(ctx); commitErr != nil {
 		sdk.Logger(ctx).Err(commitErr).Msg("teardown commit failed")
 	}
-	return s.rows.Err()
+	if rowsErr := s.rows.Err(); rowsErr != nil {
+		sdk.Logger(ctx).Err(rowsErr).Msg("rows returned an error")
+	}
+
```
dylanlott, line 147: Question: Should we return a multierror here or is logging enough?

lovromazgon: If we only log those first two errors then they are not propagated to the caller in case everything else succeeds, that's not ideal. We could use a multierror lib, although what if we simply return the first error that occurs and log any error that happens after that? Something like this:
```
...
var err error
if commitErr := s.tx.Commit(ctx); commitErr != nil {
	err = logOrReturnError(ctx, err, commitErr, "teardown commit failed")
}
...

func logOrReturnError(ctx context.Context, oldErr, newErr error, msg string) error {
	if oldErr == nil {
		return fmt.Errorf(msg+": %w", newErr)
	}
	sdk.Logger(ctx).Err(newErr).Msg(msg)
	return oldErr
}
```

Playground example: https://go.dev/play/p/lua_Yjb4avR

source/logrepl/snapshot.go
```diff
@@ -133,7 +141,15 @@ func (s *SnapshotIterator) Teardown(ctx context.Context) error {
 	if commitErr := s.tx.Commit(ctx); commitErr != nil {
```
hariso, line 141: Might be a dumb question, but why do we call s.tx.Commit() here at all?

lovromazgon: Simply to end the transaction.

hariso: Ah, I think I need a break.:D I was thinking this is the source teardown, but it's the iterator teardown, and actually a different thing in mind because of that.

source/logrepl/snapshot.go
```diff
@@ -29,6 +30,12 @@ import (
 	"github.com/jackc/pgx/v4"
 )
 
+// ErrSnapshotComplete is returned by Next when a snapshot is finished
+var ErrSnapshotComplete = errors.New("ErrSnapshotComplete")
```
lovromazgon, line 34: We could change the error message to contain something more human readable (e.g. "snapshot is completed" or similar). Same for `ErrSnapshotInterrupt`.

source/logrepl/snapshot.go
```diff
@@ -136,10 +144,20 @@ func (s *SnapshotIterator) Ack(ctx context.Context, pos sdk.Position) error {
...
-		sdk.Logger(ctx).Err(commitErr).Msg("teardown commit failed")
+		err = logOrReturnError(ctx, err, commitErr, "teardown commit failed")
+	}
+	if rowsErr := s.rows.Err(); rowsErr != nil {
+		err = logOrReturnError(ctx, err, rowsErr, "rows returned an error")
+	}
+
+	if !s.complete {
+		sdk.Logger(ctx).Warn().Msg("snapshot interrupted")
+		return ErrSnapshotInterrupt
```
lovromazgon, line 157: I think we should use the same approach as in the previous `if` statements, that is to call `logOrReturnError`.

source/logrepl/snapshot_test.go
```diff
@@ -70,7 +70,7 @@ func TestLifecycle(t *testing.T) {
 			"table":  table,
 		},
 	})
-	is.NoErr(s.Teardown(ctx)) // TODO: Should return an error
+	is.Equal(ErrSnapshotInterrupt.Error(), s.Teardown(ctx).Error())
```
lovromazgon, line 73: We should make sure the error is exactly of the type we expect using `errors.Is`.

lovromazgon (approved):
🙌

PR #31, opened 2022-04-12
[WIP] Support snapshot and CDC mode combined behavior

dylanlott:
### Description

This is an exploratory spike of the `CombinedIterator` implementation for the Postgres connector.

It should switch from snapshot to CDC mode once the snapshot is completed.

It should only operate in combined mode when `snapshotMode` is set to `initial` and `cdcMode` is set to `logRepl`.

### Quick checks:


source/logrepl/combined_iterator.go
```diff
@@ -0,0 +1,59 @@
...
+	cdc, err := NewCDCIterator(ctx, conn, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create cdc iterator: %w", err)
+	}
+
+	snap, err := NewSnapshotIterator(ctx, conn, SnapshotConfig{
+		KeyColumn:    cfg.KeyColumnName,
+		Table:        cfg.TableName,
+		Columns:      cfg.Columns,
+		SnapshotName: cdc.sub.SnapshotName,
```
lovromazgon, line 35: `sub` is an internal field, we should rather create a method that returns the snapshot name, just to keep things nicely decoupled.

source/logrepl/combined_iterator.go
```diff
@@ -0,0 +1,59 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create snapshot iterator: %w", err)
```
lovromazgon, line 38: At this point we have a running CDC iterator, we need to tear it down.

source/logrepl/combined_iterator.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// CombinedIterator binds a CDCIterator and a SnapshotIterator together for
+// coordinating the hand-off from snapshot to CDC operation modes.
+type CombinedIterator struct {
+	cdc  *CDCIterator
+	snap *SnapshotIterator
+}
+
+func NewCombinedIterator(ctx context.Context, conn *pgx.Conn, cfg Config) (*CombinedIterator, error) {
```
lovromazgon, line 23: We should probably rename the current `Config` struct to `CDCConfig` and create a separate `Config` struct for this iterator.

source/logrepl/combined_iterator.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	TypeSnapshot int = iota
+	TypeCDC
+)
+
+// CombinedIterator binds a CDCIterator and a SnapshotIterator together for
+// coordinating the hand-off from snapshot to CDC operation modes.
+type CombinedIterator struct {
```
lovromazgon, line 18: Might even be called just `Iterator`, it's the only iterator that will be used by the source so `logrepl.Iterator` looks nice IMO.

source/source.go
```diff
@@ -26,6 +26,7 @@ import (
 
 var (
 	_ Iterator = (*logrepl.CDCIterator)(nil)
```
lovromazgon, line 28: After we have this new iterator we can remove the interface check for `CDCIterator`, we don't care what it implements anymore, `Source` doesn't even have to know it exists.

PR #43, opened 2022-05-13
[WIP] Hybrid iterator

dylanlott:
### Description

**WIP** 

This PR presents a novel `HybridIterator` that can snapshot a table and transition to CDC mode. It includes feedback and addresses the same issue as #38 but was fundamentally different enough to warrant a new pull request for it's changes.

Fixes #15

### Quick checks:

- [x] I have followed the [Code Guidelines](https://github.com/ConduitIO/conduit/blob/main/docs/code_guidelines.md).
- [x] There is no other [pull request](https://github.com/ConduitIO/conduit-connector-postgres/pulls) for the same update/change.
- [ ] I have written unit tests.
- [ ] I have made sure that the PR is of reasonable size and can be easily reviewed.

PR #45, opened 2022-05-27
adds copy data writer for snapshot capabilities

dylanlott:
### Description

This PR adds the `CopyDataWriter` iterator for capturing Postgres database
snapshots. 

Previously we attempted to use a replication slot combined with Postgres'
internal snapshot capabilities, however we couldn't use `SET SNAPSHOT` keywords
because replication slots don't support the extended query protocol.

Instead, we use the `COPY` command from Postgres to dump the table in
question to stdout and parse the output into `sdk.Record` which we then
iterate.

The `CopyDataWriter` will be attached to the `Hybrid` iterator and will handle
the initial capture of a table's data within a transaction.

## Known Issues
We currently cannot handle bytea columns. We are investigating a way for the
CopyDataWriter to handle byte values differently than text values.

Dependency of #43

### Quick checks:

### DIFF
--- a/source/logrepl/copyto.go
+++ b/source/logrepl/copyto.go
@@ -32,7 +32,6 @@ import (
 
 type CopyDataWriter struct {
 	config            Config
-	conn              *pgx.Conn
 	fieldDescriptions []pgproto3.FieldDescription
 	decoders          []decoderValue
 	connInfo          *pgtype.ConnInfo
@@ -53,7 +52,6 @@ func NewCopyDataWriter(ctx context.Context, conn *pgx.Conn, config Config) (*Cop
 	}
 
 	c := &CopyDataWriter{
-		conn:              conn,
 		config:            config,
 		fieldDescriptions: fds,
 		decoders:          decoders,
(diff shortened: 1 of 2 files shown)


source/logrepl/copyto.go
```diff
@@ -0,0 +1,178 @@
...
+	copyquery := fmt.Sprintf("COPY %s TO STDOUT", c.config.TableName)
+	_, err := conn.PgConn().CopyTo(ctx, c, copyquery)
+	if err != nil {
+		fmt.Printf("failed to copy data from table: %v", err)
+	}
+	c.done <- struct{}{}
+}
+
+func (c *CopyDataWriter) Write(line []byte) (n int, err error) {
+	tokens := bytes.Split(line, []byte{'\t'})
```
lovromazgon, line 77: **Most important comment**

We need to make sure that we parse the line correctly, specifically see the documentation for [Copy - Text Format](https://www.postgresql.org/docs/current/sql-copy.html). We currently only handle the null string `\N`, but there are others (e.g. `\\`, `\b`, `\f`, `\n`, `\r`, `\t`, `\v`, `\digits`, `\xdigits`). That's also the reason why `bytea` values are not decoded correctly (the `bytea` text decoder will work once we replace `\\` with `\`). It's not as easy as just replacing values though, for example `\\n` is actually the string `\n` and not a new line. It would be almost necessary to have tests for these values to know we are handling them correctly.

source/logrepl/copyto_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}{
+		{
+			name: "should parse a copy line into a record",
+			args: args{
+				config: Config{
+					TableName: "foo",
+				},
+			},
+			setupQuery: func(conn *pgx.Conn) {
+				_, err := conn.Exec(context.Background(), `create temporary table foo(a int2, b int4, c int8, d varchar, e text, f date, g json)`)
```
lovromazgon, line 121: I'd love to see a test like the one for [`RelationSet`](https://github.com/ConduitIO/conduit-connector-postgres/blob/db6527de84dd56da9048a265b74f01a6da3de3e4/source/logrepl/internal/relationset_test.go#L43) to be sure we can handle all builtin types.

source/logrepl/copyto_test.go
```diff
@@ -0,0 +1,175 @@
...
+	<-w.done
+	is.NoErr(w.Teardown(ctx))
+}
+
+func TestCopyDataWriter_Next(t *testing.T) {
+	is := is.New(t)
+	type args struct {
+		config Config
+	}
+	tests := []struct {
```
lovromazgon, line 106: Friendly tip: don't start with a data-driven test until you know you need one. We have a single test case, this could be simplified.

lovromazgon (commented):
I posted some suggestions, but most importantly we need to make sure that the writer knows how to properly parse the copy text format.

PR #49, opened 2022-07-19
OpenCDC integration + other fixes

lovromazgon:
### Description

Updates SDK to the latest version and integrates OpenCDC record into the connector. I also took the opportunity to fix some inconsistencies, for example:
- destination test now uses global test utilities to prepare the database
- key column is included in payload in `logrepl.SnapshotIterator` to match CDC behavior and `longpoll.SnapshotIterator`
- position in `longpoll.SnapshotIterator` was changed to match position in `logrepl.SnapshotIterator`
- decoding of values was simplified and uses info from the library instead of our `pgutil` package which was now removed
- introduced constants for metadata and config fields - metadata field for table now has the prefix `postgres.*` to prevent conflicts with metadata from other connectors
- global statement builders were replaced with local ones for each struct that needs it

Code should be a bit simpler now, this can already be seen from the line number diff.

Depends on https://github.com/ConduitIO/conduit-connector-sdk/pull/31 (needs to be merged and we need to update the connector SDK version).

Related to https://github.com/ConduitIO/conduit/issues/512.

### Quick checks:


source/logrepl/snapshot.go
```diff
@@ -75,11 +71,17 @@ func NewSnapshotIterator(ctx context.Context, conn *pgx.Conn, cfg SnapshotConfig
 		return nil, fmt.Errorf("failed to load rows: %w", err)
 	}
 
+	for i, col := range cfg.Columns {
+		if col == cfg.KeyColumn {
+			s.keyColumnIndex = i
+		}
+	}
+
```
hariso, line 79: :+1:

hariso (approved):
:rocket:

PR #79, opened 2023-01-25
Add a `trigger` CDC mode

voscob:
### Description

In this pull request, a new CDC mode `trigger` has been added, instead of 'long_polling`.

### Quick checks:


source/trigger/cdc.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return fmt.Errorf("execute select query from tracking table: %w", err)
+	}
+
+	return nil
+}
+
+// newTrackingTableService is a service to delete processed rows.
+func newTrackingTableService() *trackingTableService {
+	stopCh := make(chan struct{}, 1)
```
maksenius, line 261: you can allocate variables inside trackingTableService initialization
```
return &trackingTableService{
		stopCh:     make(chan struct{}, 1),
		canCloseCh:  make(chan struct{}, 1),
		errCh:       make(chan error, 1),
		idsToDelete: make([]any, 0),
	}
```

source/trigger/position.go
```diff
@@ -0,0 +1,69 @@
...
+			Mode:      ModeSnapshot,
+			CreatedAt: time.Now().Format("150405"),
+		}, nil
+	}
+
+	if err := json.Unmarshal(position, &pos); err != nil {
+		return nil, fmt.Errorf("unmarshal sdk.Position into Position: %w", err)
+	}
+
+	return &pos, nil
```
alyakimenko, line 58: What do you think about this?
```suggestion
    pos := new(Position)
	if err := json.Unmarshal(position, pos); err != nil {
		return nil, fmt.Errorf("unmarshal sdk.Position into Position: %w", err)
	}

	return pos, nil
```
Google's style guide has a nice example of such a moment at the end of the ["Non-pointer zero values" section](https://google.github.io/styleguide/go/best-practices#non-pointer-zero-values).

source/trigger/cdc.go
```diff
@@ -0,0 +1,353 @@
...
+	"github.com/conduitio/conduit-connector-postgres/common"
+	sdk "github.com/conduitio/conduit-connector-sdk"
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"go.uber.org/multierr"
+)
+
+const (
+	timeoutBeforeCloseDBSec        = 20
+	timeoutToClearTrackingTableSec = 5
```
alyakimenko, line 35: Why not make these constants a `time.Duration` type?

```suggestion
	timeoutBeforeCloseDB        = 20 * time.Second
	timeoutToClearTrackingTable = 5 * time.Second
```

source/trigger/cdc.go
```diff
@@ -0,0 +1,353 @@
...
+	canCloseCh chan struct{}
+	// error channel
+	errCh chan error
+	// slice of identifiers to delete
+	idsToDelete []any
+}
+
+// NewCDC creates a new instance of the CDC iterator.
+func NewCDC(ctx context.Context, params CDCParams) (*CDC, error) {
+	var err error
```
alyakimenko, line 91: Do we actually need this error to be declared here?

source/trigger/iterator.go
```diff
@@ -0,0 +1,477 @@
...
+			Table:          iterator.table,
+			OrderingColumn: iterator.orderingColumn,
+			Key:            iterator.key,
+			Columns:        iterator.columns,
+			BatchSize:      iterator.batchSize,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("init snapshot iterator: %w", err)
+		}
+	} else {
```
alyakimenko, line 163: I would rather avoid the `else` by converting the `if/else` statement to a `switch-case`.

source/trigger/iterator.go
```diff
@@ -0,0 +1,477 @@
...
+}
+
+// switchToCDCIterator stops Snapshot and initializes CDC iterator.
+func (iter *Iterator) switchToCDCIterator(ctx context.Context) error {
+	err := iter.snapshot.Close()
+	if err != nil {
+		return fmt.Errorf("stop snaphot iterator: %w", err)
+	}
+
+	iter.snapshot = nil
```
alyakimenko, line 302: Wouldn't it be better to close and set the `snapshot` to `nil` _after_ the CDC is successfully initialized?

source/trigger/cdc.go
```diff
@@ -0,0 +1,353 @@
...
+		), nil
+	case actionDelete:
+		return sdk.Util.Source.NewRecordDelete(
+			convertedPosition,
+			metadata,
+			key,
+		), nil
+	default:
+		return sdk.Record{}, errWrongTrackingOperationType
+	}
```
alyakimenko, line 210: I think it would be okay to shrink the calls, as the arguments are not that long.
```suggestion
	switch operationType {
	case actionInsert:
		return sdk.Util.Source.NewRecordCreate(
		    convertedPosition, metadata, key, sdk.RawData(rowBytes),
		), nil
		
	case actionUpdate:
		return sdk.Util.Source.NewRecordUpdate(
		    convertedPosition, metadata, key, nil,sdk.RawData(rowBytes),
		), nil
		
	case actionDelete:
		return sdk.Util.Source.NewRecordDelete(
		    convertedPosition, metadata, key,
		), nil
		
	default:
		return sdk.Record{}, errWrongTrackingOperationType
	}
```

alyakimenko (changes requested):

README.md
```diff
@@ -10,42 +10,72 @@ the plugin reads from a buffer of CDC events.
...
   "logrepl.slotName": "meroxademo"
 }
 ```
 
+### Trigger Configuration
+
+The `trigger` CDC mode creates a tracking table and a trigger with a function to populate this table. The tracking
+table, trigger and a function name have the same names as a source table with a prefix of `conduit_` and a suffix of a
+creation time in `_hhmmss` format. The tracking table has all the same columns as the source table plus two additional
+columns:
```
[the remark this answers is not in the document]
voscob: Done 👍

README.md
```diff
@@ -10,42 +10,72 @@ the plugin reads from a buffer of CDC events.
...
+| `conduit_operation` | Enum operation type: `INSERT`, `UPDATE`, or `DELETE`. |
+
+Every time data is added, changed, or deleted from the source table, this event will be written to the tracking table.
+
+Queries to retrieve change data from a tracking table are very similar to queries in a Snapshot iterator, but
+with `conduit_id` ordering column.
+
+The Ack method collects the `conduit_id` of those records that have been successfully applied, in order to remove them
+later in a batch from the tracking table (every 5 seconds or when the connector is closed).
+
```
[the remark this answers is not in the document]
voscob: Done 👍

README.md
```diff
@@ -10,42 +10,72 @@ the plugin reads from a buffer of CDC events.
...
+The `trigger` CDC mode creates a tracking table and a trigger with a function to populate this table. The tracking
+table, trigger and a function name have the same names as a source table with a prefix of `conduit_` and a suffix of a
+creation time in `_hhmmss` format. The tracking table has all the same columns as the source table plus two additional
+columns:
+
+| name                | description                                           |
+|---------------------|-------------------------------------------------------|
+| `conduit_id`        | Autoincrement index for the position.                 |
+| `conduit_operation` | Enum operation type: `INSERT`, `UPDATE`, or `DELETE`. |
+
```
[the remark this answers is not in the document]
voscob: It's already there:

> The tracking table, trigger and a function name have the same names as a source table...

source/config.go
```diff
@@ -94,6 +101,9 @@ func ParseConfig(cfgRaw map[string]string) (Config, error) {
 	if cfg.Table == "" {
 		return Config{}, requiredConfigErr(ConfigKeyTable)
 	}
+	if cfg.OrderingColumn == "" {
```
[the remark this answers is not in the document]
voscob: I've decided to add this field to the order not only for the `trigger` CDC mode, but also for `logrepl`.
I think it's quite important because if sorting is not chosen, the rows will be returned in an unspecified order.
https://www.postgresql.org/docs/current/queries-order.html

source/longpoll/snapshot.go
```diff
@@ -1,206 +0,0 @@
-// Copyright © 2022 Meroxa, Inc.
```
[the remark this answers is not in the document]
voscob: Maha and I decided that we don't need a `long_polling` mode for now, because it does not support CDC.
So it has been deleted.

source/source.go
```diff
@@ -54,6 +49,12 @@ func (s *Source) Parameters() map[string]sdk.Parameter {
 			Required:    true,
 			Description: "The name of the table in Postgres that the connector should read.",
 		},
+		ConfigKeyOrderingColumn: {
+			Default:  "",
+			Required: true,
```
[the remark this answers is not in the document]
voscob: I have replied to this comment above:
https://github.com/ConduitIO/conduit-connector-postgres/pull/79#discussion_r1087045220

source/trigger/cdc.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// set a new position into the variable,
+	// to avoid saving position into the struct until we marshal the position
+	position := &Position{
+		Mode: ModeCDC,
+		// set the value from columnTrackingID column of the tracking table
+		LastProcessedVal: row[colID],
+		CreatedAt:        iter.position.CreatedAt,
```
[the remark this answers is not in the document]
voscob: Good point 👍 
I've added the comment in a `Position` struct.

source/trigger/cdc.go
voscob: Perfect 🥇 
Done.

source/source.go
```diff
@@ -72,7 +73,12 @@ func (s *Source) Parameters() map[string]sdk.Parameter {
 		ConfigKeyCDCMode: {
 			Default:     "auto",
 			Required:    false,
-			Description: "Determines the CDC mode (allowed values: `auto`, `logrepl` or `long_polling`).",
+			Description: "Determines the CDC mode (allowed values: `auto`, `logrepl` or `trigger`).",
+		},
+		ConfigKeyBatchSize: {
+			Default:     "1000",
+			Required:    false,
+			Description: "Size of rows batch. Min is 1 and max is 100000.",
```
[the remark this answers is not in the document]
voscob: My bad :(
I've updated the description.

source/trigger/position.go
```diff
@@ -0,0 +1,69 @@
...
+
+	sdk "github.com/conduitio/conduit-connector-sdk"
+)
+
+// Mode defines an iterator mode.
+type Mode string
+
+const (
+	ModeSnapshot = "snapshot"
+	ModeCDC      = "cdc"
```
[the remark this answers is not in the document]
voscob: Oh...
Fixed 👍

source/trigger/position.go
voscob: Good point 👍 
I should probably read it 🤔

source/trigger/cdc.go
voscob: Yeah, it'll be much better 👍

source/trigger/cdc.go
voscob: No, of course not, it has to be deleted.
Done.

source/trigger/iterator.go
voscob: I agree with you.
It has been simplified.

source/trigger/cdc.go
voscob: Yes, it looks much better 👍

README.md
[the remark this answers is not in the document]
maksenius: Is it one trigger for all cdc operations?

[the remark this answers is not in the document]
voscob: Yes, it has only one trigger.

source/trigger/iterator.go
alyakimenko: Yeah, I think, in this case, the order doesn't actually matter. It would just look a bit safer to me. Nevermind,and good job! :)

alyakimenko (approved):

go.mod
```diff
@@ -3,61 +3,61 @@ module github.com/conduitio/conduit-connector-postgres
...
 require (
-	github.com/Masterminds/squirrel v1.5.4
-	github.com/conduitio/conduit-connector-sdk v0.6.0
-	github.com/jackc/pgconn v1.14.0
-	github.com/jackc/pglogrepl v0.0.0-20220305000529-420b8467887a
-	github.com/jackc/pgproto3/v2 v2.3.2
-	github.com/jackc/pgtype v1.14.0
-	github.com/jackc/pgx/v4 v4.18.1
-	github.com/matryer/is v1.4.1
+    github.com/Masterminds/squirrel v1.5.4
```
lovromazgon, line 6: @simonl2002 please run `go mod tidy` to format this file in a standard way.

PR #104, opened 2023-07-26
Enable batching in destination

lovromazgon:
### Description

This PR changes the behavior for batching so that the queries for all records in a batch are sent to Postgres in one round-trip. This significantly increases performance when batching is enabled - in my tests I achieved 15k msg/s instead of previously 800.

It also updates the SDK dependency to enable the batching middleware. Before releasing the connector we should update the dependency to a proper semver version.

### Quick checks:

### DIFF
--- a/destination/destination.go
+++ b/destination/destination.go
@@ -125,7 +125,8 @@ func (d *Destination) Write(ctx context.Context, recs []sdk.Record) (int, error)
 	for i := range recs {
 		_, err := br.Exec()
 		if err != nil {
-			return i, fmt.Errorf("failed to execute query for record %d: %w", i, err)
+			// the batch is executed in a transaction, if one failed all failed
+			return 0, fmt.Errorf("failed to execute query for record %d: %w", i, err)
 		}
 	}
 	return len(recs), nil
--- a/destination/destination.go
+++ b/destination/destination.go
@@ -139,9 +139,10 @@ func (d *Destination) Teardown(ctx context.Context) error {
 	return nil
 }
 
-// handleInsert checks for the existence of a key. If no key is present it will
-// plainly insert the data. If a key exists, but no key column name is
-// configured, it attempts a plain insert to that database.
+// handleInsert adds a query to the batch that stores the record in the target
+// table. It checks for the existence of a key. If no key is present or a key
+// exists and no key column name is configured, it will plainly insert the data.
+// Otherwise it upserts the record.
 func (d *Destination) handleInsert(r sdk.Record, b *pgx.Batch) error {
 	if !d.hasKey(r) || d.config.keyColumnName == "" {
 		return d.insert(r, b)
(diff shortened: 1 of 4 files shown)


destination/destination.go
```diff
@@ -99,16 +99,34 @@ func (d *Destination) Open(ctx context.Context) error {
...
-		// TODO send all queries to postgres in one round-trip
-		err := sdk.Util.Destination.Route(ctx, r,
-			d.handleInsert,
-			d.handleUpdate,
-			d.handleDelete,
-			d.handleInsert,
-		)
+	b := &pgx.Batch{}
+	for _, rec := range recs {
+		var err error
```
maha-hajja, line 104: we should move this line outside the loop 👀

destination/destination.go
```diff
@@ -99,16 +99,34 @@ func (d *Destination) Open(ctx context.Context) error {
...
+		}
+	}
+
+	br := d.conn.SendBatch(ctx, b)
+	defer br.Close()
+
+	for i := range recs {
+		_, err := br.Exec()
+		if err != nil {
+			// the batch is executed in a transaction, if one failed all failed
```
maha-hajja, line 128: maybe we should add this info to the error msg too? to show that more than one record failed to be written

destination/destination.go
```diff
@@ -124,30 +142,30 @@ func (d *Destination) Teardown(ctx context.Context) error {
 // handleInsert checks for the existence of a key. If no key is present it will
 // plainly insert the data. If a key exists, but no key column name is
 // configured, it attempts a plain insert to that database.
-func (d *Destination) handleInsert(ctx context.Context, r sdk.Record) error {
+func (d *Destination) handleInsert(r sdk.Record, b *pgx.Batch) error {
```
maha-hajja, line 145: IIUC these methods don't actually execute the query anymore, they create the query and add it to a queue, so we should update the docs for each method accordingly

maha-hajja (approved):
LGTM, nice work with this huge performance gap!

destination/destination.go
lovromazgon: Why though? 🤔 This error is only relevant inside the for loop.

destination/destination.go
lovromazgon: We return 0 so the SDK will nack all records in the batch, so the user will know that all failed. The error message will just point to the specific record that caused the batch to fail.

destination/destination.go
maha-hajja: hmm my reasoning was to not create a new variable each time in the loop, and because when err is not nil we will return from the method, so no need to have a new empty err each time

lovromazgon: I expect the Go compiler to optimize this. You made me curious so I created two simple Go programs, the only difference between them being the place where the variable is declared. The compiler actually produced the same assembly code for both programs. Of course, I can't say _for sure_ this is the case here, but I sure expect that it is.

That said, _even if_ the produced assembly code is different, I'd be surprised if it made a noticeable difference. And given that we're writing code for humans and not for machines, I put more importance on writing clear and understandable code, which includes limiting the scope of variables to avoid accidental reuse.

maha-hajja: oh interesting! thank you for taking this extra step and checking the assembly code.
yeah I agree it's definitely more understandable too

maha-hajja (approved):
nice work!

PR #105, opened 2023-07-26
Cleanup

lovromazgon:
### Description

This cleanup refactors the code by putting the source and destination structs in the root of the repository and changes the config parsing to use paramgen.

### Quick checks:

- [ ] I have written unit tests. (no behavioral change, existing unit tests pass)


destination_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2022 Meroxa, Inc.
+// Copyright © 2023 Meroxa, Inc.
```
maha-hajja, line 1: I'm not sure if I remember right, but didn't we agree on leaving the original copyright year when updating a file? not a big deal ofc just asking

source.go
```diff
@@ -12,87 +12,71 @@
...
+	URL string `json:"url" validate:"required"`
+	// The name of the table in Postgres that the connector should read.
+	Table string `json:"table" validate:"required"`
+	// Comma separated list of column names that should be included in each Record's payload.
+	Columns []string `json:"columns"`
+	// Column name that records should use for their `Key` fields.
+	Key string `json:"key"`
+
+	// Whether or not the plugin will take a snapshot of the entire table before starting cdc mode.
+	SnapshotMode source.SnapshotMode `json:"snapshotMode" validate:"inclusion=initial|never" default:"initial"`
```
maha-hajja, line 50: I know this is out of the scope and a bit too late to change, but can't understand why we didn't just go with a boolean here 😅

source.go
```diff
@@ -12,87 +12,71 @@
...
-			Required:    false,
-			Description: "Name of the publication to listen for WAL events.",
-		},
-		ConfigKeyLogreplSlotName: {
-			Default:     "conduitslot",
-			Required:    false,
-			Description: "Name of the slot opened for replication events.",
-		},
-	}
+	return nil
```
maha-hajja, line 67: what why :O

source/config.go
```diff
@@ -14,44 +14,6 @@
 
 package source
```
maha-hajja, line 15: maybe we should rename the file? it doesn't have any config info anymore

destination.go
```diff
@@ -58,37 +58,24 @@ func NewDestination() sdk.Destination {
...
-			Required:    false,
-			Description: "The name of the table in Postgres that the connector should write to.",
-		},
-		ConfigKey: {
-			Default:     "",
-			Required:    false,
-			Description: "Column name used to detect if the target table already contains the record.",
-		},
-	}
+	return DestinationConfig{}.Parameters()
```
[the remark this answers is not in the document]
lovromazgon: Not really, I can change it to `d.config.Parameters()` if that's what we use elsewhere 🙂

destination_test.go
lovromazgon: It was done automatically by the IDE, I think it thought it's a new file and updated the license. I'll revert it 👍

source.go
lovromazgon: I think the idea was to give us the freedom to add more modes in the future, e.g. [debezium has 6](https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-connector-snapshot-mode-options).

source.go
lovromazgon: 😱 Good catch!

source/config.go
lovromazgon: How would you name it? 🙃

maha-hajja: hmmm `modes`? `modeTypes`? maybe even move the types to another file?
just because we usually have the config struct under this name, and it's usually the first file I check, so it's a bit confusing

lovromazgon: I couldn't find a good name, it got me thinking something is fishy. I realized that we can move `SourceConfig` into `source` and have a `source.Config` instead, then it also made sense to have these types in the `config.go` file as they are types related to fields in the config. I think it's cleaner now 🙂 I made the same changes for the destination config.

maha-hajja (approved):
LGTM

PR #113, opened 2023-12-18
support reading from multiple tables using logical replication slots

### LINKED ISSUE #23: Feature: Support multiple tables
### Feature description

It would be great if the connector exposed the support for more than one table at a time. At the moment, the number of tables dictates the number of replication slots which is not ideal.

maha-hajja:
### Description
  
support reading from multiple tables using logical replication slots.

changes to params:
* "columns" param is removed.
* "key" param is updates now to accept multiple tables and keys, as a list of pairs, example: "table1:key1,table2:key2", if the key is not provided, we'll fetch the table's primary key.

Fixes #23 

### Quick checks:

### DIFF
--- a/source.go
+++ b/source.go
@@ -77,7 +77,7 @@ func (s *Source) Open(ctx context.Context, pos sdk.Position) error {
 			Position:        pos,
 			SlotName:        s.config.LogreplSlotName,
 			PublicationName: s.config.LogreplPublicationName,
-			TableName:       s.config.Table,
+			Tables:          s.config.Table,
 			KeyColumnName:   s.config.Key,
 			Columns:         s.config.Columns,
 		})
@@ -96,7 +96,7 @@ func (s *Source) Open(ctx context.Context, pos sdk.Position) error {
 		snap, err := longpoll.NewSnapshotIterator(
 			ctx,
 			s.conn,
-			s.config.Table,
+			s.config.Table[0], //todo: only the first table for now
 			s.config.Columns,
 			s.config.Key)
 		if err != nil {
--- a/source.go
+++ b/source.go
@@ -64,6 +64,7 @@ func (s *Source) Open(ctx context.Context, pos sdk.Position) error {
 		return fmt.Errorf("failed to connect to database: %w", err)
 	}
 	s.conn = conn
+	fmt.Println("COLUMNS: ", columns)
 
 	switch s.config.CDCMode {
 	case source.CDCModeAuto:
@@ -136,7 +137,7 @@ func (s *Source) Teardown(ctx context.Context) error {
 }
 
 func (s *Source) getTableColumns(ctx context.Context, conn *pgx.Conn) ([]string, error) {
-	query := "SELECT column_name FROM information_schema.columns WHERE table_name = ?"
+	query := "SELECT column_name FROM information_schema.columns WHERE table_name = $1"
 
 	rows, err := conn.Query(ctx, query, s.config.Table[0])
 	if err != nil {
--- a/source.go
+++ b/source.go
@@ -74,7 +74,7 @@ func (s *Source) Open(ctx context.Context, pos sdk.Position) error {
 	if err != nil {
 		return fmt.Errorf("failed to connect to database: %w", err)
 	}
-	columns, err := s.getTableColumns(conn)
+	columns, err := s.getTableColumns(ctx, conn)
 	if err != nil {
 		return fmt.Errorf("failed to connect to database: %w", err)
 	}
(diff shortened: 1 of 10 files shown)


source/paramgen.go
```diff
@@ -19,13 +19,13 @@ func (Config) Parameters() map[string]sdk.Parameter {
 		},
 		"columns": {
 			Default:     "",
-			Description: "Comma separated list of column names that should be included in each Record's payload.",
+			Description: "todo: reove param, Comma separated list of column names that should be included in each Record's payload.",
```
lovromazgon, line 22: The todo comment shows up here. Not a biggie, if we're going to remove this before we release the next version.

source/logrepl/cdc.go
```diff
@@ -154,21 +154,26 @@ func (i *CDCIterator) attachSubscription(ctx context.Context, conn *pgx.Conn) er
...
-	keyColumn, err := i.getKeyColumn(ctx, conn)
-	if err != nil {
-		return fmt.Errorf("failed to find key for table %s (try specifying it manually): %w", i.config.TableName, err)
+	var err error
+	keyColumnMp := make(map[string]string, len(i.config.Tables))
+	for _, tableName := range i.config.Tables {
+		// Call function and store the result in the map
+		keyColumnMp[tableName], err = i.getKeyColumn(ctx, conn, tableName)
+		if err != nil {
+			return fmt.Errorf("failed to find key for table %s (try specifying it manually): %w", tableName, err)
```
lovromazgon, line 163: IIUC we plan to remove the key column, so the error message is a bit misleading. Although, thinking about it, we could accept comma separated pairs, where each pair is separated by a colon, e.g.:
```yaml
table: "employees,users,addresses",
key: "employees:emp_id,users:id",
```
And only if we don't find a manually specified key column, we fall back to fetching it from the DB (like for `addresses` in the example above).

lovromazgon (approved):
LGTM, there's a few followups we need to do, as you already pointed out in comments. I'll collect them here:

- update the snapshot iterator to fetch data from multiple tables
- remove `key` and `column` config fields
- think about actually having a `key` field that accepts comma separated `table:column` pairs
- later on introduce a Conduit-wide convention for specifying the source table/topic/collection that can be used for routing in destination connectors and add the metadata field instead of (or in addition to?) the `postgres.table` metadata field

I think at least the first three would be needed to make this feature complete so we can cut a release.

source/paramgen.go
maha-hajja: this was intentional cuz it won't be merged until the params are removed

source.go
```diff
@@ -131,3 +149,25 @@ func (s *Source) Teardown(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (s *Source) getTableColumns(conn *pgx.Conn) ([]string, error) {
+	query := fmt.Sprintf("SELECT column_name FROM information_schema.columns WHERE table_name = '%s'", s.config.Table[0])
```
lovromazgon, line 154: SQL injection alert ‼️ Please inject user input as a query parameter.

source.go
```diff
@@ -131,3 +149,25 @@ func (s *Source) Teardown(ctx context.Context) error {
...
+
+	rows, err := conn.Query(context.Background(), query)
```
lovromazgon, line 156: We have a nice context that we get from `Open`, please use that, in case the query for some reason takes longer than expected and Conduit wants to abort it.

source.go
```diff
@@ -131,3 +149,25 @@ func (s *Source) Teardown(ctx context.Context) error {
...
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var columns []string
+	for rows.Next() {
```
lovromazgon, line 163: We need to check `rows.Err()` after the loop, see [handling errors](http://go-database-sql.org/errors.html).

source.go
maha-hajja: oh that was definitely unintentional, thanks for noticing it!

source/logrepl/cdc.go
```diff
@@ -37,8 +37,7 @@ type Config struct {
 	SlotName        string
 	PublicationName string
 	Tables          []string
-	KeyColumnName   string
-	Columns         []string
+	KeyColumnMp     map[string]string
```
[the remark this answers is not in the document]
maha-hajja: yes agreed

source/config_test.go
```diff
@@ -0,0 +1,79 @@
...
+package source
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/matryer/is"
+)
+
+func TestConfig_Validate(t *testing.T) {
```
lovromazgon, line 24: Nice 👍

lovromazgon (approved):

PR #129, opened 2024-04-08
feat(cdc): read from multiple tables

raulb:
### Description

Part of https://github.com/ConduitIO/conduit/issues/1473

- [x] Update source configuration to read from `opencdc.collection` instead of from `postgres.table` in logrepl.cdc.
- [x] Implement wildcard functionality to listen to all tables (`*`)
- [x] Update readme for both source and destination (part of https://github.com/ConduitIO/conduit-connector-postgres/pull/131)

### Quick checks:


README.md
```diff
@@ -69,13 +69,13 @@ handle different payloads and keys. Because of this, each record is individually
 
 ## Table Name
 
-If a record contains a `postgres.table` property in its metadata it will be inserted in that table, otherwise it will
+If a record contains an `opencdc.collection` property in its metadata it will be inserted in that table, otherwise it will
```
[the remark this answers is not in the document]
raulb: Oops, didn't mean to update the destination bits on this pull-request. Thanks for pointing that out.

source.go
```diff
@@ -158,3 +167,30 @@ func (s *Source) getTableColumns(ctx context.Context, conn *pgx.Conn) ([]string,
 	}
 	return columns, nil
 }
+
+func (s *Source) readingAllTables() bool {
+	return len(s.config.Table) == 1 && s.config.Table[0] == source.AllTablesWildcard
+}
+
+func (s *Source) getAllTables(ctx context.Context, conn *pgx.Conn) ([]string, error) {
+	query := "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public' AND table_type = 'BASE TABLE'"
```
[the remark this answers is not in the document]
raulb: @lovromazgon I was thinking about making this base on the public schema and then create a separate pull-request for letting the user specifying the schema, making `public` as default.

source.go
```diff
@@ -63,6 +64,14 @@ func (s *Source) Open(ctx context.Context, pos sdk.Position) error {
...
 	}
+
+	if s.readingAllTables() {
+		s.config.Table, err = s.getAllTables(ctx, conn)
+		if err != nil {
+			return fmt.Errorf("failed to connect to database: %w", err)
+		}
+	}
+
 	s.conn = conn
```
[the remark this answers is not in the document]
raulb: @lovromazgon good catch, thank yoy

.github/workflows/validate-generated-files.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Check generated files
+        run: |
+          export PATH=$PATH:$(go env GOPATH)/bin
+          make install-tools generate
+          git diff --exit-code --numstat
```
hariso, line 23: Nitpick: might be good to print the diff as well.

raulb: @hariso i see you shipped this one on https://github.com/ConduitIO/conduit-connector-postgres/pull/131/files#diff-721a78cbc4708b9650ff1a84a8dcbc1a5c166b5a9c2e0b1f648a952c9e21a35d, so I'll rebase.

README.md
```diff
@@ -52,31 +52,21 @@ can't be determined it will fail.
...
-
+| name                      | description                                                                                                                                                                                | required | default       |
+|---------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|---------------|
+| `url`                     | Connection string for the Postgres database.                                                                                                                                               | true     |               |
+| `tables`                  | List of table names to read from, separated by comma. Example: `"employees,offices,payments"`. Using `*` will read from all public tables.                                                 | true     |               |
```
lovromazgon, line 58: Btw, if we're going to make this change in a separate PR we could also move the readme changes there.

README.md
```diff
@@ -52,31 +52,21 @@ can't be determined it will fail.
...
+| `key`                     | List of Key column names per table, separated by comma. Example:`"table1:key1,table2:key2"`, if not supplied, the table(s) primary keys will be used as the `'Key'` field for the records. | false    |               |
```
hariso, line 59: i.e. this assumes one key per table?

source.go
```diff
@@ -123,39 +147,86 @@ func (s *Source) Ack(ctx context.Context, pos sdk.Position) error {
...
+// table if one exists and returns it.
+func (s *Source) getTableKeys(ctx context.Context, tableName string) (string, error) {
+	query := `SELECT c.column_name
+FROM information_schema.table_constraints tc
+JOIN information_schema.constraint_column_usage AS ccu USING (constraint_schema, constraint_name)
+JOIN information_schema.columns AS c ON c.table_schema = tc.constraint_schema
+  AND tc.table_name = c.table_name AND ccu.column_name = c.column_name
+WHERE constraint_type = 'PRIMARY KEY' and tc.table_name = $1`
+
+	rows, err := s.connPool.Query(ctx, query, tableName)
```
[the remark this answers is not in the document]
lyuboxa: See https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS

source/logrepl/cdc.go
```diff
@@ -52,13 +59,19 @@ type CDCIterator struct {
...
 // until either the context is canceled or Teardown is called.
-func NewCDCIterator(ctx context.Context, conn *pgx.Conn, config Config) (*CDCIterator, error) {
+func NewCDCIterator(ctx context.Context, connPool *pgxpool.Pool, config Config) (*CDCIterator, error) {
 	i := &CDCIterator{
 		config:  config,
 		records: make(chan sdk.Record),
 	}
 
-	err := i.attachSubscription(ctx, conn)
+	conn, err := connPool.Acquire(ctx)
```
lyuboxa, line 68: I'd rather this be avoided, rather use the pool directly in the iterator for the following reasons:
1. The CDC iterator uses a single replication connection always
2. No need to explicitly release the connection, so safer for leaks.

lyuboxa: Wait, actually I just realized the way this flow is a bit odd. Since the pg repl connection doesn't actually need the pool but just the config of the connection, since the pool does not support replication connection types. 

So the only thing that will be extracted is the config, which makes the Acquire/Release an overhead since that connection  won't be used.

source/snapshot/iterator.go
```diff
@@ -83,14 +85,19 @@ func (i *Iterator) Next(ctx context.Context) (sdk.Record, error) {
 			if err := i.t.Err(); err != nil {
 				return sdk.Record{}, fmt.Errorf("fetchers exited unexpectedly: %w", err)
 			}
+			if err := i.acks.Wait(ctx); err != nil {
```
lyuboxa, line 88: Curious what are we guarding against here? Potentially Conduit calling next without saving the previous record?

README.md
raulb: Will address on https://github.com/ConduitIO/conduit-connector-postgres/pull/135. However, as discussed today at stand up, we're removing `key` altogether since we're going to rely on primary keys automatically.

source.go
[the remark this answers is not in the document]
lovromazgon: You can create a composite key and this query will return multiple rows in that case. The issue is that the connector won't know what to do with multiple keys right now, so we want to fail early.

source/logrepl/cdc.go
lovromazgon: We can supply the pool down into `attachSubscription` but we'll have to acquire a connection there anyway 🤷 We need a `TypeMap()` in line 179, as far as I can see we can only get it from a `pgx.Conn`. Note that this connection is released right away and is not actually used to fetch any data.

source/snapshot/iterator.go
lovromazgon: We're guarding against shutting down the pipeline before all records are processed. Since we don't have long polling right now, the pipeline will stop with an error saying `snapshot complete`, but we don't want that error to be emitted until we received all acks, otherwise some records might get dropped.

source/logrepl/cdc.go
lyuboxa: No worries, to note is that establishing connection in Postgres is pretty expensive, since it needs to allocate all the workmem.

lovromazgon: Yeah, but do you see a way around it? We need that type map.

source.go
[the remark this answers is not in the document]
lyuboxa: Indeed, this is an excellent point.

source/logrepl/cdc.go
lyuboxa: I believe this is just `pgtype.NewMap()` which contains all the types.

lyuboxa: Though, I'm not sure how this may work with custom types which are not known by the map

source.go
```diff
@@ -123,39 +147,86 @@ func (s *Source) Ack(ctx context.Context, pos sdk.Position) error {
...
+}
+
+func (s *Source) readingAllTables() bool {
+	return len(s.config.Table) == 1 && s.config.Table[0] == source.AllTablesWildcard
 }
 
-func (s *Source) getTableColumns(ctx context.Context, conn *pgx.Conn) ([]string, error) {
-	query := "SELECT column_name FROM information_schema.columns WHERE table_name = $1"
+func (s *Source) getAllTables(ctx context.Context) ([]string, error) {
+	query := "SELECT table_name FROM information_schema.tables WHERE table_schema = 'public' AND table_type = 'BASE TABLE'"
```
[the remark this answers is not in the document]
lovromazgon: So I looked into this and I'm not sure I'm following. `pg_tables` doesn't contain any interesting info, we'd still need to join all the tables we already have in the query, what do we gain by also joining the `pg_tables` view here?

source/logrepl/cdc.go
lyuboxa: According by this https://pkg.go.dev/github.com/jackc/pgtype, new types will need to be explicitly registered so we are ok.

source.go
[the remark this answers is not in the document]
lovromazgon: I noticed I didn't specify the schema, so I'll add `AND tc.table_schema = 'public'`.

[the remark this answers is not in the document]
lyuboxa: What interesting info? This query only selects the table name, which is available in `pg_tables`.

source/logrepl/cdc.go
lovromazgon: Ok using `pgtype.NewMap` now 👍

source.go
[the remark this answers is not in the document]
lovromazgon: Ohh, I was looking at the primary key query 🤦 my bad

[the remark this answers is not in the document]
lovromazgon: Done 👍

PR #132, opened 2024-04-16
New snapshot iterator and position format

lyuboxa:
### Description

The iterator supports snapshots based on range and is limited to tables which have an ordinal number (integer) as primary key. If these conditions are not met, validation will fail.

Support for multiple tables is builtin. Each table gets a dedicated worker. Positional data keeps track of the progress of all tables, not just the one which the record is related to.

Resuming a snapshot is supported, if the snapshot positional data requires it.

This work is not integrated into the CDC or LongPoll iterator and the new position type is not compatible. This will work will follow.

Fixes # (issue)

### Quick checks:


source/snapshot/iterator.go
```diff
@@ -0,0 +1,182 @@
...
+	p, err := position.ParseSDKPosition(c.Position)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse position: %w", err)
+	}
+
+	if p.Snapshot == nil {
+		p.Snapshot = make(position.SnapshotPositions)
+	}
+
+	db, err := pgxpool.New(ctx, conninfo)
```
lovromazgon, line 60: Why are we creating a separate connection pool? 😢 We should let the caller give us a connection. My reasoning is that when doing long polling, we can get away with a single connection pool, we only need a separate connection if we do logical replication, in which case the caller (i.e. the combined iterator) should take care that it opens a separate connection and provides it here.

lyuboxa: Fair point, I did this to the fetchers but not so much in the iterator. Fixed in d4c8a39

The pool only keeps connection which were created via the `.Acquire()` method. Any other querier methods will do the acquire and release automagically. With Postgres only a single command per connection can be processed at the time. The fetchers each use a long running tx to pull the data, which then will automatically be released.

source/position/position_test.go
```diff
@@ -0,0 +1,96 @@
...
+				panicErr = errors.New(fmt.Sprint(r))
+			}
+		}()
+
+		var p *Position
+		_ = p.ToSDKPosition()
+	}()
+
+	is.True(panicErr != nil)
+	is.Equal(panicErr.Error(), "runtime error: invalid memory address or nil pointer dereference")
```
hariso, line 57: Hmm, what exactly is being tested with this? It reads like a generic test when accessing a nil pointer, nothing position specific?

source/snapshot/iterator.go
```diff
@@ -0,0 +1,173 @@
...
+			defer wg.Done()
+
+			if err := f.Run(ctx); err != nil {
+				return fmt.Errorf("fetcher exited: %w", err)
+			}
+
+			return nil
+		})
+	}
+	defer close(i.records)
```
hariso, line 170: When fetchers terminate, we might still have some records in this channel. WDYT about draining the channel before actually calling it done?

lyuboxa: Does that need to happen?  My thinking is this:

1. Channel is not buffered, so fetchers will wait on injecting records as long as next is called.
2. If fetchers are done, then there is no more data coming through the channel, which means they are not blocking on sending data
3. Once all fetchers have exited the channel can be closed, since should be empty.

Now if the channel is buffered, Next() should be able to drain the channel  it hits the not ok when receiving. 
Even if there are any messages in the channel and they get dropped because of teardown or cancelation, the position will not be updated which means we can safely resume from when the last position was saved.

Does that make sense?

source/snapshot/fetcher.go
```diff
@@ -0,0 +1,382 @@
...
+			f.conf.Table: {
+				LastRead:    lastRead,
+				SnapshotEnd: f.snapshotEnd,
+				Done:        f.snapshotEnd == lastRead,
+			},
+		},
+	}.ToSDKPosition()
+
+	meta := map[string]string{
+		"table": f.conf.Table,
```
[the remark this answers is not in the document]
lyuboxa: Yea, I left that for @raulb to tackle but can make the change.

source/snapshot/fetcher.go
```diff
@@ -0,0 +1,382 @@
...
+		f.conf.Key,
+	)
+
+	if _, err := tx.Exec(ctx, cursorSQL); err != nil {
+		return nil, err
+	}
+
+	return func() {
+		// N.B. The cursor will automatically close when the TX is done.
+		if _, err := tx.Exec(ctx, fmt.Sprint("CLOSE ", f.cursorName)); err != nil {
```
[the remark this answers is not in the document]
lyuboxa: @hariso I can, so many ways to combine strings it nauseating .. ;-)

source/position/position.go
```diff
@@ -0,0 +1,82 @@
...
+
+const (
+	TypeInitial Type = iota
+	TypeSnapshot
+	TypeCDC
+)
+
+type Position struct {
+	Type     Type              `json:"type"`
+	Snapshot SnapshotPositions `json:"snapshot,omitempty"`
```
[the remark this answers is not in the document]
lyuboxa: yea, I went back and forth between these, since a snapshot can contain N tables, I decided to pluralize the positions part, as in multiple positions before multiple tables. what do you think?

source/position/position.go
```diff
@@ -0,0 +1,82 @@
...
+func (p Position) ToSDKPosition() sdk.Position {
+	v, err := json.Marshal(p)
+	if err != nil {
+		panic(err)
+	}
+
+	return sdk.Position(v)
+}
+
+func (p Position) LSN() (pglogrepl.LSN, error) {
```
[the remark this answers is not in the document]
lyuboxa: Will add this on all parts, thanks

source/position/position_test.go
lyuboxa: `ToSDKPosition` does not return an error but panics when marshal fails, so not super easy to test in the regular part, just a test guard.

source/snapshot/fetcher.go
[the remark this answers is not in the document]
lyuboxa: Fixed in 4423c91

source/snapshot/fetcher.go
```diff
@@ -0,0 +1,382 @@
...
+		ctx,
+		fmt.Sprintf("SET TRANSACTION SNAPSHOT '%s'", f.conf.Snapshot),
+	); err != nil {
+		return fmt.Errorf("failed to set tx snapshot %q: %w", f.conf.Snapshot, err)
+	}
+
+	return nil
+}
+
+func validateKey(ctx context.Context, table, key string, tx pgx.Tx) error {
```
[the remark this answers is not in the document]
lyuboxa: Fixed in 4423c91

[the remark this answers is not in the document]
lyuboxa: @hariso regarding this part:

> Maybe also add checks the key's type is integer?

Yea, this is missing right now because I haven't figured an elegant way of doing it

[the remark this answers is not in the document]
lyuboxa: @hariso this is as elegant as it gets, see dcc4816

source/position/position_test.go
lovromazgon: > NaN and +/-Inf values will return an [UnsupportedValueError](https://pkg.go.dev/encoding/json#UnsupportedValueError).

and

> Channel, complex, and function values cannot be encoded in JSON. Attempting to encode such a value causes Marshal to return an [UnsupportedTypeError](https://pkg.go.dev/encoding/json#UnsupportedTypeError).

So I think we're safe, we don't have any of those in the position.

source/snapshot/fetcher.go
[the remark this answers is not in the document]
lovromazgon: 😈 
```
strings.ReplaceAll(errors.Join(errors.New("CURSOR "), errors.New(f.cursorName)).Error(), "\n", "")
```

Makefile
```diff
@@ -5,21 +5,33 @@ build:
 	go build -ldflags "-X 'github.com/conduitio/conduit-connector-postgres.version=${VERSION}'" -o conduit-connector-postgres cmd/connector/main.go
 
 .PHONY: test
-test:
-	# run required docker containers, execute integration tests, stop containers after tests
-	docker compose -f test/docker-compose.yml up --quiet-pull -d --wait
-	go test $(GOTEST_FLAGS) -race ./...; ret=$$?; \
-		docker compose -f test/docker-compose.yml down; \
-		exit $$ret
+test: start_test_service tests stop_test_service
```
lovromazgon, line 8: This will keep the docker containers running if tests fail 😢

source/config_test.go
```diff
@@ -25,43 +25,44 @@ func TestConfig_Validate(t *testing.T) {
...
 		wantErr bool
-	}{{
-		name: "valid config",
-		cfg: Config{
-			URL:     "postgresql://meroxauser:meroxapass@127.0.0.1:5432/meroxadb",
-			Table:   []string{"table1", "table2"},
-			Key:     []string{"table1:key1"},
-			CDCMode: CDCModeLogrepl,
+	}{
+		{
```
lovromazgon, line 29: Any specific reason you modified this formatting? Seems like the linter doesn't complain about it and neither do the formatting tools..

source/position/position_test.go
lyuboxa: @lovromazgon remove the test in this case?

Makefile
lyuboxa: Yea, this is the downside, however running `make tests` each time you need a test checked was a little bit much.
Since nearly all tests require running Postgres. 

In all honesty, the makefile have no business managing docker compose.. rather it should just bail if the URL is not provided. 

If we do want to make this easier to clean up - perhaps we can shove it in `t.Main` and add a cleanup hook?

source/config_test.go
lyuboxa: ah, because of gofumpt...

source/snapshot/iterator.go
```diff
@@ -0,0 +1,173 @@
...
+	"gopkg.in/tomb.v2"
+)
+
+var ErrIteratorDone = errors.New("snapshot complete")
+
+type Config struct {
+	Position   sdk.Position
+	Tables     []string
+	TablesKeys map[string]string
+	Snapshot   string
```
lovromazgon, line 35: This spawned a slight WTF moment in my head, I realized it's the snapshot ID. WDYT about renaming this (and similar fields) to `SnapshotID`?

source/snapshot/iterator.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	if len(errs) != 0 {
+		return errors.Join(errs...)
+	}
+
+	return nil
+}
+
+func (i *Iterator) start(ctx context.Context) {
```
lovromazgon, line 153: Nitpick: I'd name this `run`, since it blocks, also it will match the same name in the worker.

source/snapshot/iterator.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		i.workers[j] = w
+	}
+
+	if len(errs) != 0 {
+		return errors.Join(errs...)
+	}
+
+	return nil
```
lovromazgon, line 150: The function does this check for us.
```suggestion
	errors.Join(errs...)
```

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+		ctx,
+		fmt.Sprintf("SET TRANSACTION SNAPSHOT '%s'", f.conf.Snapshot),
+	); err != nil {
+		return fmt.Errorf("failed to set tx snapshot %q: %w", f.conf.Snapshot, err)
+	}
+
+	return nil
+}
+
+func (FetchWorker) validateKey(ctx context.Context, table, key string, tx pgx.Tx) error {
```
lovromazgon, line 339: Let's not mix pointer and value receivers please 🙏 Same for `validateTable`.

```suggestion
func (*FetchWorker) validateKey(ctx context.Context, table, key string, tx pgx.Tx) error {
```

source/position/position_test.go
lovromazgon: If you want, yes, I think we can safely remove it.

Makefile
lovromazgon: The thing is that we use this pattern in all connectors and I prefer to have consistency rather than creating a special way to start/stop containers here. It adds to the cognitive load when working on different repos. So unless it's an important improvement and we want to port it to other connectors as well, I'd rather leave it as it was.

source/snapshot/iterator.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, fmt.Errorf("failed to parse position: %w", err)
+	}
+
+	if p.Snapshot == nil {
+		p.Snapshot = make(position.SnapshotPositions)
+	}
+
+	i := &Iterator{
+		db:           db,
+		t:            &tomb.Tomb{},
```
lovromazgon, line 62: We could just pass the context to the tomb using `tomb.WithContext`, then we know that anything started by the tomb will have access to the correct context.

source/snapshot/iterator.go
```diff
@@ -0,0 +1,173 @@
...
+			if err := f.Run(ctx); err != nil {
+				return fmt.Errorf("fetcher exited: %w", err)
+			}
+
+			return nil
+		})
+	}
+	defer close(i.records)
+
+	wg.Wait()
```
lovromazgon, line 172: Isn't this the same as running `<-i.t.Dead()`? We don't need the wait group then.

source/snapshot/iterator.go
```diff
@@ -0,0 +1,173 @@
...
+func (i *Iterator) Teardown(_ context.Context) error {
+	if i.t != nil {
+		i.t.Kill(errors.New("tearing down snapshot iterator"))
+	}
+
+	return nil
+}
+
+func (i *Iterator) updateLastPosition(r *sdk.Record) error {
+	pos, err := position.ParseSDKPosition(r.Position)
```
lovromazgon, line 110: It's a bit of a shame that the worker marshals the position just so it's unmarshaled here again. It's not the end of the world though 🙃

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+
+	if t, ok := c.Position.Snapshot[c.Table]; ok {
+		f.snapshotEnd = t.SnapshotEnd
+		f.lastRead = t.LastRead
+	}
+
+	return f
+}
+
+// Validate will ensure the config is correctt.
```
lovromazgon, line 111:
```suggestion
// Validate will ensure the config is correct.
```

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+
+func (f *FetchWorker) Run(ctx context.Context) error {
+	start := time.Now().UTC()
+
+	tx, err := f.db.BeginTx(ctx, pgx.TxOptions{
+		IsoLevel:   pgx.RepeatableRead,
+		AccessMode: pgx.ReadOnly,
+	})
+	if err != nil {
+		return err
```
lovromazgon, line 150: We should probably wrap the error and say we failed to start a transaction.

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+		// N.B. The cursor will automatically close when the TX is done.
+		if _, err := tx.Exec(ctx, "CLOSE "+f.cursorName); err != nil {
+			sdk.Logger(ctx).Warn().
+				Err(err).
+				Msgf("unexpected error when closing cursor %q", f.cursorName)
+		}
+	}, nil
+}
+
+func (f *FetchWorker) updateFetchLimit(ctx context.Context, tx pgx.Tx) error {
```
lovromazgon, line 228: Nitpick: I'd either rename this method to `updateSnapshotEnd` or rename the field to `fetchLimit`, so the function name and the field it's updating match.

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+func (f *FetchWorker) createCursor(ctx context.Context, tx pgx.Tx) (func(), error) {
+	cursorSQL := fmt.Sprintf("DECLARE %s CURSOR FOR (SELECT * FROM %s WHERE %s > %d AND %s <= %d ORDER BY %s)",
+		f.cursorName,
+		f.conf.Table,
+		f.conf.Key,
+		f.lastRead,
+		f.conf.Key,
+		f.snapshotEnd,
+		f.conf.Key,
+	)
```
lovromazgon, line 212: It smells like SQL injection in here 👃 I wonder if we can make this safer, at least for the values that are provided by the user (key and table, right?).

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+
+	tx, err := f.db.BeginTx(ctx, pgx.TxOptions{
+		IsoLevel:   pgx.RepeatableRead,
+		AccessMode: pgx.ReadOnly,
+	})
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := tx.Rollback(ctx); err != nil {
```
lovromazgon, line 153: Just thinking out loud - is a rollback appropriate here? Would it make sense to rather commit? It's not like we are changing anything. Or rollback in case we are returning an error and commit otherwise. Does it make a difference?

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+		values, err := rows.Values()
+		if err != nil {
+			return 0, fmt.Errorf("failed to get values")
+		}
+
+		if err := f.send(
+			ctx,
+			f.buildRecord(fields, values),
+		); err != nil {
+			return nread, fmt.Errorf("failed to send record")
```
lovromazgon, line 267: Whoops, we are losing the error here.

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+	for _, f := range rows.FieldDescriptions() {
+		fields = append(fields, f.Name)
+	}
+
+	var nread int
+
+	for rows.Next() {
+		values, err := rows.Values()
+		if err != nil {
+			return 0, fmt.Errorf("failed to get values")
```
lovromazgon, line 260: Here as well.

source/snapshot/fetch_worker.go
```diff
@@ -0,0 +1,393 @@
...
+		switch t := values[i].(type) {
+		case time.Time: // type not supported in sdk.Record
+			payload[name] = t.UTC().String()
+		default:
+			payload[name] = t
+		}
+	}
+
+	// always coerce snapshot position to bigint, pk may be any type of integer.
+	lastRead := keyInt64(payload[f.conf.Key])
```
lovromazgon, line 298: Are we confident this will never panic? Otherwise I'd rather return an error.

source/snapshot/iterator.go
lyuboxa: Sure, I can also call it `TxSnapshotID` to make it even more explicit

source/snapshot/fetch_worker.go
lyuboxa: Oooh sure, will do.

source/config_test.go
lovromazgon: Yeah I figured, no problem, just curious 😉

source/snapshot/fetch_worker.go
lyuboxa: `Table`, `Key` are provided by the user and validation happens before that. SQL can be injected if the validation passes somehow, but the queries there are prepared so it should fail.

source/snapshot/fetch_worker.go
lyuboxa: The tx will be discarded anyway, so it doesn't matter. We can commit, I just don't see a reason why.
Cursors require an active tx. Since we are not changing anything didn't make sense. I will double check

source/snapshot/fetch_worker.go
lyuboxa: Well, actually, it probably will, since the types we are allow are also real/double and numeric. 

I think the ideal situation is to drop all non-integers, and then this will be fine. Having pk to be numeric or real is something odd...

source/snapshot/fetch_worker.go
lovromazgon: Don't waste time on this, I was just wondering if there's a case to be made for one or the other. Probably it won't make a difference either way.

Makefile
lyuboxa: can revert it no problem.

source/position/position_test.go
```diff
@@ -0,0 +1,96 @@
...
+	is := is.New(t)
+
+	invalid := Position{LastLSN: "invalid"}
+	_, err := invalid.LSN()
+	is.True(err != nil)
+	is.Equal(err.Error(), "failed to parse LSN in position: failed to parse LSN: expected integer")
+
+	valid := Position{LastLSN: "4/137515E8"}
+	_, noErr := valid.LSN()
+	is.NoErr(noErr)
```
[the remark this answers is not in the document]
lyuboxa: I can add this, but this is testing more LSN parsing, which this code does not do.

source/snapshot/fetch_worker.go
lyuboxa: Had another look at the [docs](https://www.postgresql.org/docs/current/sql-declare.html)

TL;DR it only matters if `WITH HOLD` is specified which will persist the cursor when the transaction completes,
in which case it makes sense to commit rather than rollback.

As an aside, in some situations `WITH HOLD` can be used with the following considerations.

1. Declare cursor with hold and it exists outside of the TX 
2. FETCH in parallel from multiple connections, but this can screw up ordering if one is fetching faster.
3. Cursor will need to be closed explicitly in this case.

One beneficial use to using this option is if we don't want to maintain the transaction open and this may become very reasonable if locks are causing contention, in which case the flow will be something like this:

1. Generate consistent cursor name
2. Declare cursor with hold
3. Fetch in a single connection 
4. Close cursor when done

Resuming becomes more complicated, because we will either need to close and redeclare the cursor (based in the name provided in 1), or figure out current position of cursor and seek back, but this will require the `SCROLL`  option.

I'm happy to leave it as is, until we need to put more work.

lyuboxa:
@lovromazgon addressed (I think) most of your comments here 2d60841

source/snapshot/fetcher.go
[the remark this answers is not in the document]
hariso: @lyuboxa Looks good!

source/position/position.go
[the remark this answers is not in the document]
raulb: @lyuboxa Oh, I meant the variable name (not the type). I'm guessing that maybe `Snapshot` is still more appropriate, right?

source/snapshot/fetch_worker.go
lovromazgon: I changed this to return an error so we at least don't panic right now, in case we encounter this.

source/snapshot/iterator.go
lovromazgon: For the record - I changed this so the fetcher returns the position as a Go object alongside the payload and key, the iterator then stitches it together into a `sdk.Record`.

source/snapshot/iterator.go
hariso: Yup, makes sense, I missed that the channel was unbuffered, that helped a lot.:)

source/position/position.go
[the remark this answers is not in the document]
lovromazgon: I added a comment above 👍

source/snapshot/iterator.go
lovromazgon: Sorry @hariso, didn't reload the page when I posted this comment. GitHub should really implement live loading of new comments, this happened a few times already 😞

source/position/position.go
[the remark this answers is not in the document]
lovromazgon: I renamed it here https://github.com/ConduitIO/conduit-connector-postgres/pull/132/commits/ebdec835649ab533e1a52115b21c7c027754a20d.

hariso (approved):
LGTM! I guess the docs (README) will be updated once the changes have been integrated?

lovromazgon:
@hariso yeah I think we can postpone that, since these changes don't have any effect yet.

raulb (approved):

PR #139, opened 2024-04-29
Add default logrepl autoCleanup mode to remove replication slots and publications

lyuboxa:
### Description

The new option `logrepl.autoCleanup` is not only valid in the context of logical replication. The default is set to true and will remove the replication slot and publication when the connector is deleted.

This change is a departure from the current implementation, in the following areas:

1. Replication slots are not `temporary` any longer and will persist through pipeline state changes.
2. An implication of 1) the slot will need to be cleaned up, hence why the `autoCleanup` option is set to true.

In cases where the replication slots and publications are pre-created, ensure `logrepl.autoCleanup` is set to `false`.

Fixes # (issue)

### Quick checks:

### DIFF
--- a/source/config.go
+++ b/source/config.go
@@ -68,6 +68,10 @@ type Config struct {
 	// LogreplSlotName determines the replication slot name in case the
 	// connector uses logical replication to listen to changes (see CDCMode).
 	LogreplSlotName string `json:"logrepl.slotName" default:"conduitslot"`
+
+	// LogreplAutoCleanup determines if the replication slot and publication should be
+	// removed when the connector is deleted.
+	LogreplAutoCleanup bool `json:"logrepl.autoCleanup" validate:"inclusion=true|t,exclusion=false|f" default:"true"`
 }
 
 // Validate validates the provided config values.
--- a/test/helper.go
+++ b/test/helper.go
@@ -99,6 +99,27 @@ func SetupTestTable(ctx context.Context, t *testing.T, conn Querier) string {
 	return table
 }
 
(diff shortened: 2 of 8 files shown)


test/helper.go
```diff
@@ -99,6 +99,27 @@ func SetupTestTable(ctx context.Context, t *testing.T, conn Querier) string {
...
+	is := is.New(t)
+
+	_, err := conn.Exec(
+		context.Background(),
+		"SELECT pg_create_logical_replication_slot($1, $2)",
+		slotName,
+		"pgoutput",
+	)
+
+	is.NoErr(err)
```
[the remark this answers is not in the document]
lyuboxa: @lovromazgon done, see 38d7f17

This also includes a note in the README. I found few other cases where the slot is not cleaned up and fix these too.

source/config.go
```diff
@@ -68,6 +68,10 @@ type Config struct {
 	// LogreplSlotName determines the replication slot name in case the
 	// connector uses logical replication to listen to changes (see CDCMode).
 	LogreplSlotName string `json:"logrepl.slotName" default:"conduitslot"`
+
+	// LogreplAutoCleanup determines if the replication slot and publication should be
+	// removed when the connector is deleted.
+	LogreplAutoCleanup bool `json:"logrepl.autoCleanup" validate:"inclusion=true|t,exclusion=false|f" default:"true"`
```
hariso, line 74: `exclusion` would mean that the value cannot be set to `false` or `f`, which I guess we don't want?

lyuboxa: yea since this is a bool we can remove the validation entirely ¯\_(ツ)_/¯

PR #141, opened 2024-05-06
Add snapshot support to logrepl mode

### LINKED ISSUE #15: Snapshot support in logrepl mode

lyuboxa:
### Description

The new combined iterator integrates snapshots with the existing cdc iterator.
Once the snapshot is completed, it will swithc to cdc.

When using snapshots, the connector will run both cdc and snapshot at the same time.
This is required because the cdc iterator needs to be initialized and export a consistent
snapshot which is taken at the time of the slot creation. The cdc iteator will need to
keep the connection open until the snapshot is done.

The switch to the cdc iterator is triggered when the snapshot returns an end of records error.
Based on this signal, the cdc subscriber will be started to consume changes from the replication slot.

Fixes #15 
### Quick checks:


source/logrepl/cdc.go
```diff
@@ -23,72 +23,100 @@ import (
...
 }
 
 // NewCDCIterator sets up the subscription to a logical replication slot and
 // starts a goroutine that listens to events. The goroutine will keep running
 // until either the context is canceled or Teardown is called.
-func NewCDCIterator(ctx context.Context, connPool *pgxpool.Pool, config Config) (*CDCIterator, error) {
-	i := &CDCIterator{
-		config:  config,
-		records: make(chan sdk.Record),
+func NewCDCIterator(ctx context.Context, pgconf *pgconn.Config, c CDCConfig) (*CDCIterator, error) {
```
lyuboxa, line 62: We can pass a connection here, but it needs to be established with the `replication=true` param, which will enable replication protocol commands in the same connection. If we do this, the connection will need to be opened in `source.go`. However is that connection is only used in the CDC iterator, it can be established here for now.

Alternatively, we can move it one step higher to the combined iterator. Either way, this is going to be an integration test.

source/logrepl/combined.go
```diff
@@ -0,0 +1,177 @@
...
+		SlotName:        conf.SlotName,
+		PublicationName: conf.PublicationName,
+		Tables:          conf.Tables,
+		TableKeys:       conf.TableKeys,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create CDC iterator: %w", err)
+	}
+
+	c := &CombinedIterator{
```
hariso, line 80: IIUC, the logic should be: if `WithSnapshot` is true, and the position type is `Snapshot`, that means that the snapshot is required, and it's not done, so we need to use the snapshot iterator. In all other cases (snapshot not required, snapshot required but done), we activate the CDC iterator. Is that correct?

source/logrepl/combined.go
```diff
@@ -0,0 +1,177 @@
...
+	c.activeIterator = snapshotIterator
+
+	return nil
+}
+
+func (c *CombinedIterator) useCDCIterator(ctx context.Context) (sdk.Record, error) {
+	sdk.Logger(ctx).Debug().
+		Msg("snapshot done, switching to CDC mode")
+
+	c.activeIterator, c.snapshotIterator = c.cdcIterator, nil
```
hariso, line 168: Maybe we should tear down the snapshot iterator here?

hariso (commented):
Great work, Lyu!

source/logrepl/cdc.go
```diff
@@ -23,72 +23,100 @@ import (
...
 }
 
-func (c Config) LSN() (pglogrepl.LSN, error) {
-	if len(c.Position) == 0 {
-		return 0, nil
-	}
-	lsn, err := PositionToLSN(c.Position)
-	if err != nil {
-		return 0, fmt.Errorf("failed to parse position: %w", err)
+func (c Config) Validate() error {
```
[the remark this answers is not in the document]
lyuboxa: Yup, will add one.

source/logrepl/cdc.go
```diff
@@ -23,72 +23,100 @@ import (
...
 	}
 
-	err := i.attachSubscription(connPool.Config().ConnConfig.Config)
-	if err != nil {
-		return nil, fmt.Errorf("failed to setup subscription: %w", err)
+	if err := internal.CreatePublication(
+		ctx,
+		conn,
+		c.PublicationName,
+		internal.CreatePublicationOptions{Tables: c.Tables},
```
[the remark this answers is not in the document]
lyuboxa: Um, this needs to be fixed, since `*` is all tables. Thanks for raising this.

source/logrepl/cdc.go
```diff
@@ -23,72 +23,100 @@ import (
...
+		records: records,
+		pgconn:  conn,
+		sub:     sub,
+	}, nil
 }
 
 // listen should be called in a goroutine. It starts the subscription and keeps
 // it running until the subscription is stopped or the context is canceled.
-func (i *CDCIterator) listen(ctx context.Context) {
+func (i *CDCIterator) Start(ctx context.Context) error {
```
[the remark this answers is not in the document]
lyuboxa: Will do.

source/logrepl/combined.go
lyuboxa: `WithSnapshot` is the config value representing that this connector should take a snapshot, so the value won't change. Thus there are two cases:

* With Snapshot enabled
  * The position can be unknown or initial - use snapshot iterator
  * The position can be snapshot (to resume snapshot) - use snapshot iterator
  * The position can be CDC (snapshot completed and CDC has taken over) - use cdc iterator
* With snapshot disabled
  * It is only CDC mode


Does that make sense?

source/logrepl/combined.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, fmt.Errorf("failed to create CDC iterator: %w", err)
+	}
+
+	c := &CombinedIterator{
+		cdcIterator:    cdcIterator,
+		activeIterator: cdcIterator,
+	}
+
+	// Resume with completed snapshot and CDC streaming
+	if conf.WithSnapshot && pos.Type == position.TypeCDC {
```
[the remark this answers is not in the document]
lyuboxa: No, we don't thanks for catching that.

source/logrepl/combined.go
lyuboxa: We can, but we will need to re-create it when new tables are added, this is the safer thing to do. Will get it done.

source/logrepl/combined_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/conduitio/conduit-connector-postgres/test"
+	// sdk "github.com/conduitio/conduit-connector-sdk"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/matryer/is"
+)
+
+func Test_Combined(t *testing.T) {
+	// ctx, cancel := context.WithTimeout(context.Background(), time.Second * 20)
+	// defer cancel()
```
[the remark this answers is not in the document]
lyuboxa: This is just an experiment, I will have it cleaned up a bit later

source/logrepl/internal/subscription_test.go
```diff
@@ -14,6 +14,7 @@
 
 package internal
 
+/*
```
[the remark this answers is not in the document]
lyuboxa: Yea, I will need to fix it ;-)

source/logrepl/internal/subscription.go
```diff
@@ -27,125 +27,119 @@ import (
...
+	ctx context.Context,
+	conn *pgconn.PgConn,
 	slotName,
 	publication string,
 	tables []string,
 	startLSN pglogrepl.LSN,
 	h Handler,
-) *Subscription {
+) (*Subscription, error) {
+	result, err := pglogrepl.CreateReplicationSlot(
```
lovromazgon, line 65: One thing that caught my eye is that `NewSubscription`, a constructor, has a side effect other than creating the `Subscription` object. If you ask me, it's a code smell.

source/logrepl/combined.go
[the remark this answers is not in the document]
lovromazgon: Maybe this is what we want?
```suggestion
	if !conf.WithSnapshot || pos.Type == position.TypeCDC {
```

source/logrepl/cdc.go
```diff
@@ -23,72 +23,100 @@ import (
...
+func NewCDCIterator(ctx context.Context, pgconf *pgconn.Config, c CDCConfig) (*CDCIterator, error) {
+	conn, err := pgconn.ConnectConfig(ctx, withReplication(pgconf))
+	if err != nil {
+		return nil, fmt.Errorf("could not establish replication connection: %w", err)
 	}
 
-	err := i.attachSubscription(connPool.Config().ConnConfig.Config)
-	if err != nil {
-		return nil, fmt.Errorf("failed to setup subscription: %w", err)
+	if err := internal.CreatePublication(
```
lovromazgon, line 68: I don't get it, what's the reason behind pushing this logic from `internal.Subscription` into the CDC iterator?

source/logrepl/combined.go
lovromazgon: I also had to check the logic a few times to understand it, the rate of [WTFs/minute](https://i0.wp.com/commadot.com/wp-content/uploads/2009/02/wtf.png) was a bit higher here, maybe it would be worth refactoring to make this flow a bit clearer.

source/logrepl/combined.go
lovromazgon: What @hariso said ☝️

source/logrepl/cdc.go
lyuboxa: I don't think it belongs there, `publication` and `subscriptions` are two different objects. Albeit a sub requiring a pub, shouldn't manage it. Since `subscription` only responsibility is read CDC events. 

The CDC iterator in turn can manage the publication.

source/logrepl/combined.go
lyuboxa: Yea, I will. It is somewhat annoying because in all cases the CDC iterator needs to be created. The snapshot iterator in this turn is optional in few cases.

source/logrepl/combined.go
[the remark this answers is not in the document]
lyuboxa: Yea this is what I have it refactored to

source/logrepl/combined.go
hariso: I agree with Lovro, it would be really good if the code would follow the cases you mentioned more closely. Plus, if possible, removing the `initial` type, since it's synonymous with position being `nil`.

source/logrepl/internal/subscription.go
```diff
@@ -462,3 +346,35 @@ func (s *Subscription) receiveMessage(ctx context.Context, conn *pgconn.PgConn,
 	}
 	return msg, nil
 }
+
+// doneReplication performs the replication closing tasks on completition and
+// closes the done channel. If any errors are encountered, will be available through Err().
+func (s *Subscription) doneReplication() {
+	fmt.Println("done replication")
```
hariso, line 353: You know, logs, fmt.Println...:)

source/logrepl/combined.go
lyuboxa: I refactored this a bit here, 7d40dad

It removes a bit of the mental load around initialization.
* The CDC iterator needs to always be initialized (and maybe started)
* The Snapshot iterator only needs to be initialized when snapshotting is enabled and CDC was not used last.

> Plus, if possible, removing the initial type, since it's synonymous with position being nil.

The initial position designates no state. Since the position cannot be nil, since it is a value type, we need to assign some meaning to the default value, this is why initial exists. Not sure how to communicate this in another way, besides using ptr or explicitly checking for the default value.

source/logrepl/internal/subscription.go
```diff
@@ -27,125 +27,119 @@ import (
...
+
+	<-s.ready
+
+	return nil
 }
 
 // listen runs until context is cancelled or an error is encountered.
-func (s *Subscription) listen(ctx context.Context, conn *pgconn.PgConn) error {
+func (s *Subscription) listen(ctx context.Context) error {
+	defer s.doneReplication()
```
hariso, line 128: It might help if you give `ctx` to `s.doneReplication()`, since that method uses `context.Background()`.

lyuboxa: this was on purpose, so replication clean up cannot be cancelled, otherwise it will be half done.

source/logrepl/combined.go
lyuboxa: this was done

source/logrepl/combined.go
```diff
@@ -0,0 +1,214 @@
...
+	if err := c.initCDCIterator(ctx, pos); err != nil {
+		return nil, err
+	}
+
+	// Initialize the snapshot iterator when snapshotting is enabled and not completed.
+	if conf.WithSnapshot && pos.Type != position.TypeCDC {
+		if err := c.initSnapshotIterator(ctx); err != nil {
+			return nil, err
+		}
+	}
```
lovromazgon, line 84: One more suggestion to make this easier to read:
1. I'd move the condition into `initSnapshotIterator`, let the function decide when a snapshot iterator is needed.
2. I'd add an explicit step _after_ initializing both iterators, where we choose the active iterator. It can be as simple as checking if the snapshot iterator exists.

```suggestion
	if err := c.initSnapshotIterator(ctx, pos); err != nil {
		return nil, fmt.Errorf("failed to initialize snapshot iterator: %w", err)
	}
	
	// choose active iterator
	switch {
	case c.snapshotIterator != nil:
		c.activeIterator = c.snapshotIterator
	default:
		c.activeIterator = c.cdcIterator
		if err := c.cdcIterator.Start(ctx); err != nil {
			return fmt.Errorf("failed to start CDC iterator: %w", err)
		}
	}
```

lyuboxa: yea good point, originally, I had the condition inside the init but decide to move it to be explicit.

source/logrepl/combined.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil, fmt.Errorf("failed to validate longrepl config: %w", err)
+	}
+
+	c := &CombinedIterator{
+		conf: conf,
+		pool: pool,
+	}
+
+	// Initialize the CDC iterator. If snapshotting is disabled or position is of
+	// CDC type, the iterator will be started as well.
```
lovromazgon, line 74:
```suggestion
	// Initialize the CDC iterator.
```

source/logrepl/cdc.go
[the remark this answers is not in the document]
lyuboxa: This was removed here aa5536c . Seems like the option is not used since we iterate over all tables and save them in the config.

source/logrepl/combined.go
lyuboxa: Added

source/logrepl/cdc.go
```diff
@@ -18,83 +18,117 @@ import (
...
+
+	return nil
 }
 
 // Next returns the next record retrieved from the subscription. This call will
 // block until either a record is returned from the subscription, the
 // subscription stops because of an error or the context gets canceled.
+// Returns error when the subscription has been started.
 func (i *CDCIterator) Next(ctx context.Context) (sdk.Record, error) {
+	if started := i.subStarted.Load(); !started {
```
hariso, line 128: Maybe `!i.subStarted.Load()`?

lyuboxa: sorry I changed this code, I can simplify it.

lyuboxa: fixed here f8b0599

source/logrepl/internal/subscription_test.go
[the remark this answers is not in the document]
lyuboxa: These tests were refactored and enabled.

source/logrepl/cdc.go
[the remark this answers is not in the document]
lyuboxa: Renamed `Start` to `Run` since the function will block. See 5d83fd8

source/logrepl/internal/subscription.go
lyuboxa: cleaned up

source/logrepl/combined_test.go
```diff
@@ -0,0 +1,272 @@
...
+		TableKeys: map[string]string{
+			"t1": "k1", "t4": "k4",
+		},
+	}.Validate()
+
+	errs := err.(interface{ Unwrap() []error }).Unwrap()
+
+	is.True(len(errs) == 2)
+	is.Equal(errs[0].Error(), `missing key for table "t2"`)
+	is.Equal(errs[1].Error(), `missing key for table "t3"`)
```
hariso, line 46: Nitpick: I _think_ this should work too and simplifies the code a bit:
```go
	is.Equal(
		err,
		errors.Join(errors.New(`missing key for table "t2"`), errors.New(`missing key for table "t3"`)),
	)
```

source/logrepl/combined_test.go
```diff
@@ -0,0 +1,272 @@
...
+			Tables:          []string{table},
+			TableKeys:       map[string]string{table: "id"},
+			PublicationName: table,
+			SlotName:        table,
+			WithSnapshot:    false,
+		})
+		is.NoErr(err)
+
+		is.True(i.snapshotIterator == nil)
+		is.True(i.cdcIterator == i.activeIterator)
```
hariso, line 77: Maybe also check that it's not nil?

hariso (approved):
LGTM! I left a couple of non-blockers. Great work, Lyubo!

source/logrepl/combined_test.go
lyuboxa: Yea that is easier, thanks!

source/logrepl/internal/subscription.go
```diff
@@ -451,14 +329,46 @@ func (s *Subscription) sendStandbyStatusUpdate(ctx context.Context, conn *pgconn
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to receive message: %w", err)
 	}
 	return msg, nil
 }
+
+// doneReplication performs the replication closing tasks on completition and
+// closes the done channel. If any errors are encountered, will be available through Err().
+func (s *Subscription) doneReplication() {
+	if err := s.sentStandbyDone(context.Background()); err != nil {
```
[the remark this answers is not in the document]
lyuboxa: @hariso how about a timeout context, but not the parent?

source/logrepl/cleaner.go
```diff
@@ -52,6 +53,14 @@ func Cleanup(ctx context.Context, c CleanupConfig) error {
 	var errs []error
 
 	if c.SlotName != "" {
+		// Terminate any outstanding backends which are consuming the slot before deleting it.
+		mrr := conn.Exec(ctx, fmt.Sprintf(
+			"SELECT pg_terminate_backend(active_pid) FROM pg_replication_slots WHERE slot_name='%s AND active=true'", c.SlotName,
```
lovromazgon, line 58: Hmm, aren't the quotes misplaced here?

```suggestion
			"SELECT pg_terminate_backend(active_pid) FROM pg_replication_slots WHERE slot_name='%s' AND active='true'", c.SlotName,
```

source/logrepl/internal/publication_test.go
```diff
@@ -94,59 +99,6 @@ func TestCreatePublicationForTables(t *testing.T) {
...
-			AllTables: true,
-		},
-	)
-
-	is.True(err != nil)
-	_, ok := err.(*pgconn.PgError)
-	is.True(!ok) // err shouldn't be a PgError, no statement should be even sent to Postgres
-}
-
-func TestCreatePublicationAllTables(t *testing.T) {
```
lovromazgon, line 118: Did we remove support for creating a publication for all tables? The connector config [advertises](https://github.com/ConduitIO/conduit-connector-postgres/blob/10e4575daf336ec9bea04dd6970e2f429667d901/source/config.go#L55) this functionality.

source/logrepl/internal/subscription.go
```diff
@@ -451,14 +329,46 @@ func (s *Subscription) sendStandbyStatusUpdate(ctx context.Context, conn *pgconn
...
+
+// sentStandbyDone signals replication done and submits the last flushed LSN.
+func (s *Subscription) sentStandbyDone(ctx context.Context) error {
+	var errs []error
+
+	// send copy done message indicating replication is done
+	if err := s.sendStandbyCopyDone(ctx); err != nil {
+		sdk.Logger(ctx).Error().
+			Err(err).
+			Msg("failed to seend standby copy done")
```
lovromazgon, line 362:
```suggestion
			Msg("failed to send standby copy done")
```

test/helper.go
```diff
@@ -62,6 +62,17 @@ func ConnectSimple(ctx context.Context, t *testing.T, connString string) *pgx.Co
 	return conn.Conn()
 }
 
+func ConnectReplication(ctx context.Context, t *testing.T, connString string) *pgconn.PgConn {
+	is := is.New(t)
+
+	conn, err := pgconn.Connect(ctx, connString+"&replication=database")
+	is.NoErr(err)
+	t.Cleanup(func() {
+		is.NoErr(conn.Close(ctx))
```
lovromazgon, line 71: The test context might get cancelled, we still want to execute `Close` using an open context in that case.
```suggestion
		is.NoErr(conn.Close(context.Background()))
```

source/logrepl/cleaner.go
lyuboxa: Yes, but just on `slot_name` in pg SQL `true` does not need to be quoted.

source/logrepl/internal/publication_test.go
lyuboxa: The refactor which happened earlier, which reads all tables, does not communicate to the iterator that it should create a publication on all tables, but rather it provides a slice of tables (as per getAllTables). 

The solution for this will be to update the publication with the new table list. As per this https://github.com/ConduitIO/conduit-connector-postgres/issues/145

source/logrepl/combined_test.go
lyuboxa: Fixed here f28285404c22064f4ac63af063fcc785358965c3

source/logrepl/internal/subscription.go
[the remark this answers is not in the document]
lyuboxa: Fixed here f28285404c22064f4ac63af063fcc785358965c3

source/logrepl/cleaner.go
lyuboxa: Fixed here f28285404c22064f4ac63af063fcc785358965c3

source/logrepl/cdc.go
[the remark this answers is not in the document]
lyuboxa: test is in here f28285404c22064f4ac63af063fcc785358965c3

source/logrepl/combined.go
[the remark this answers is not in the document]
lyuboxa: This was refactored

source/logrepl/internal/subscription.go
lyuboxa: see f28285404c22064f4ac63af063fcc785358965c3, timeout added

source/logrepl/internal/publication_test.go
lovromazgon: I see, so we manually fetch a list of all tables and populate the config field. All good 👍

source/logrepl/cdc.go
```diff
@@ -176,7 +181,15 @@ func (i *CDCIterator) Ack(_ context.Context, sdkPos sdk.Position) error {
 // or the context gets canceled. If the subscription stopped with an unexpected
 // error, the error is returned.
 func (i *CDCIterator) Teardown(ctx context.Context) error {
-	defer i.pgconn.Close(ctx)
+	defer func() {
+		// Give the subscription 2 seconds to settle.
+		select {
+		case <-i.sub.Done():
+		case <-time.After(subscriberDoneTimeout):
+		}
```
lovromazgon, line 189: This part is doing the same as `i.sub.Wait` just with an additional 2 second delay. Why don't we just change `Subscription.Wait()` to allow us to do this? For example:

```go
i.sub.Wait(ctx, subscriberDoneTimeout)
```

and

```go
func (s *Subscription) Wait(ctx context.Context, minimumWaitDelay time.Duration) error {
	select {
	case <-s.done:
		return nil
	case <-time.After(minimumWaitDelay):
		// Delay reached, continue waiting for as long as the context is open.
	}
	select {
	case <-ctx.Done():
		return ctx.Err()
	case <-s.done:
		return nil
	}
}
```

lovromazgon (approved):
Looks good, I added one last suggestion. Good work 👍

Issue #151, opened 2024-05-21
Remove CDC mode long_polling from configs and docs


hariso:
### Feature description

Currently, it's not possible to use long polling in CDC. It was left as a placeholder in the configuration and the docs, but never implemented. We should remove it, because it's confusing to users.

The implementation of long polling CDC will be tracked here: https://github.com/ConduitIO/conduit-connector-postgres/issues/150.

PR #153, opened 2024-05-23
feat: remove long_polling

raulb:
### Description

Fixes https://github.com/ConduitIO/conduit-connector-postgres/issues/151

### Quick checks:


README.md
```diff
@@ -57,16 +57,16 @@ the connector will return an error.
...
+|---------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|----------|---------------|
+| `url`                     | Connection string for the Postgres database.                                                                                                  | true     |               |
+| `tables`                  | List of table names to read from, separated by comma. Example: `"employees,offices,payments"`. Using `*` will read from all public tables.    | true     |               |
+| `snapshotMode`            | Whether or not the plugin will take a snapshot of the entire table before starting cdc mode (allowed values: `initial` or `never`).           | false    | `initial`     |
+| `cdcMode`                 | Determines the CDC mode (allowed values: `auto`, `logrepl`).                                                                                  | false    | `auto`        |
```
hariso, line 65: Since we have only `logrepl` available, it looks like `cdcMode` is not needed anymore.

raulb: @hariso that's my thought as well. However, do we want to keep it since we know we'll implement it?

hariso: Hmm, I'd say it's the same situation as with long_polling: having it as a placeholder could be confusing.

raulb: @hariso @lovromazgon my thinking is to _signal_ users that there will be more modes to come beyond what's default rather than to indicate this only will be meant for logical replication.

hariso (approved):

PR #154, opened 2024-05-24
Fix operation for snapshots and replace use of information_schema

lyuboxa:
### Description

Snapshot records were being produced using the `Create` operation rather than 'Snapshot'. PostgreSQL `information_schema` can lack visibility to users with limited grants, switch to using pg_* tables where possible.

Use jackc pgerrcode instead of maintaining our own constants.

### Quick checks:


anna-cross (approved):

samirketema (approved):

raulb (approved):
LGTM. Thank you @lyuboxa !

PR #155, opened 2024-05-24
Ensure default is used when cleanup is not explicitly provided

lyuboxa:
### Description

This fixes an issue where the publication and replication slot were not removed on teardown.
The config needs to be properly initialized, defaults including.

Current sdk, does not enrich the config when passed to lifecycle events. This PR handles this on the side.

### Quick checks:


lovromazgon (approved):
We'll need to update the code when upgrading to the new version of the SDK, as it will expose functions for sanitizing the config and apply defaults.

PR #161, opened 2024-05-30
Handle numeric type in cdc and snapshot

### LINKED ISSUE #160: Bug: CDC/Snapshot cannot handle numeric type
### Bug description

When a table has a `numeric` type, the connector will fail to protobuf encode the value, since it is not a plugin supported type.  The connector will need to coerce the value to its proper kind.

```
2024-05-29T13:27:35+00:00 ERR pipeline stopped error="node pg-to-file:pg.in stopped with error: source stream was stopped unexpectedly: error reading from source: read stream error: error converting payload: error converting after: proto: invalid type: pgtype.Numeric
```

### Steps to reproduce

1. Create table with `numerc(12,2)` type and load some data.
2. Start a pipeline with a pg connector for the specific table. 
3. Should see error in description 

### Version

latest

lyuboxa:
### Description

`Numeric` or `pgtype.Numeric` is a complex type which is not supported by the plugin transport and will need to be reduced to its simpler type. In this case depending on the scale it is either a float or an integer. 

If the exponent type is negative, the Postgres numeric type is likely of the format `NUMERIC(x,y)`, conversely when the exponent is positive or equal to zero it is like defined as `NUMERIC(x,0)` or simply `NUMERIC(x)`.  

Thus, a numeric type with precision is coerced to its float64 value and without precision to an int64.

Fixes #160 

### Quick checks:


source/types/numeric.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+type NumericFormatter struct {
+	Value pgtype.Numeric
+}
+
+// Format coerces `pgtype.Numeric` to int or double depending on the exponent.
+// Returns error when value is invalid.
+func (n NumericFormatter) Format() (any, error) {
```
hariso, line 27: Nitpick: if the function should always return an int or a float, maybe we can use generics to indicate that. Also, if time allows, would be good to add some unit tests for this.

hariso (approved):
Great and quick work! :muscle:

hariso (approved):

source/types/numeric.go
lyuboxa: Fixed in last refactor.

PR #167, opened 2024-06-10
Improve handling of keepalive messages and manage server end lag.

lyuboxa:
### Description

* Ensure expected commit lsn matches that of begin messages.
* Adjust the status update to send the last server lsn when all data nas been written and acked. This change will help with managing server lag with relatively less frequently updated tables.

### Quick checks:

### DIFF
--- a/source/logrepl/internal/subscription.go
+++ b/source/logrepl/internal/subscription.go
@@ -351,7 +351,7 @@ func (s *Subscription) sendStandbyStatusUpdate(ctx context.Context) error {
 
 	// N.B. Manage replication slot lag, by responding with the last server LSN, when
 	//      all previous slot relevant msgs have been written and flushed
-	if walFlushed == s.walWritten && s.walFlushed < s.serverWALEnd {
+	if walFlushed == s.walWritten && walFlushed < s.serverWALEnd {
 		if err := pglogrepl.SendStandbyStatusUpdate(ctx, s.conn, pglogrepl.StandbyStatusUpdate{
 			WALWritePosition: s.serverWALEnd,
 			ClientTime:       time.Now(),
--- a/source/logrepl/internal/subscription.go
+++ b/source/logrepl/internal/subscription.go
@@ -95,7 +95,8 @@ func CreateSubscription(
 	slotLSN, err = pglogrepl.ParseLSN(result.ConsistentPoint)
 	if err != nil {
 		if startLSN == 0 {
-			return nil, fmt.Errorf("failed to parse consistent wal lsn: %w", err)
+			sdk.Logger(ctx).Warn().
+				Msg("start LSN is zero, using existing replication slot without position")
 		}
 
 		slotLSN = startLSN
@@ -230,6 +231,12 @@ func (s *Subscription) handleXLogData(ctx context.Context, copyDataMsg *pgproto3
 	writtenLSN, err := s.Handler(ctx, logicalMsg, xld.WALStart)
 	if err != nil {
 		return fmt.Errorf("handler error: %w", err)
+	}
+
+	// When starting on an existing slot without having a consistent slot LSN,
+	// set the flushed WAL LSN to just before the received LSN
+	if s.walWritten == 0 && s.ConsistentSlotLSN == 0 {
+		atomic.StoreUint64((*uint64)(&s.walFlushed), uint64(writtenLSN-1))
 	}
 
 	if writtenLSN > 0 {
--- a/source/logrepl/internal/subscription.go
+++ b/source/logrepl/internal/subscription.go
@@ -160,13 +160,11 @@ func (s *Subscription) listen(ctx context.Context) error {
 
 		switch copyDataMsg.Data[0] {
 		case pglogrepl.PrimaryKeepaliveMessageByteID:
-			err := s.handlePrimaryKeepaliveMessage(ctx, copyDataMsg)
-			if err != nil {
+			if err := s.handlePrimaryKeepaliveMessage(ctx, copyDataMsg); err != nil {
 				return err
 			}
 		case pglogrepl.XLogDataByteID:
-			err := s.handleXLogData(ctx, copyDataMsg)
-			if err != nil {
+			if err := s.handleXLogData(ctx, copyDataMsg); err != nil {
 				return err
 			}
 		default:
@@ -187,7 +185,7 @@ func (s *Subscription) handlePrimaryKeepaliveMessage(ctx context.Context, copyDa
 		return fmt.Errorf("failed to parse primary keepalive message: %w", err)
 	}
 
-	s.serverWALEnd = pkm.ServerWALEnd
+	atomic.StoreUint64((*uint64)(&s.serverWALEnd), uint64(pkm.ServerWALEnd))
 
 	if pkm.ReplyRequested {
 		if err := s.sendStandbyStatusUpdate(ctx); err != nil {
--- a/source/logrepl/internal/subscription.go
+++ b/source/logrepl/internal/subscription.go
@@ -34,13 +34,14 @@ const (
 
 // Subscription manages a subscription to a logical replication slot.
 type Subscription struct {
-	SlotName      string
-	Publication   string
-	Tables        []string
-	StartLSN      pglogrepl.LSN
-	Handler       Handler
-	StatusTimeout time.Duration
-	TXSnapshotID  string
+	SlotName          string
+	Publication       string
+	Tables            []string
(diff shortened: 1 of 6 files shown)


source/logrepl/internal/subscription.go
```diff
@@ -214,9 +218,31 @@ func (s *Subscription) handleXLogData(ctx context.Context, copyDataMsg *pgproto3
...
+			Stringer("tx_lsn", m.TransactionEndLSN).
+			Stringer("expected_end_lsn", s.lastTXLSN).
+			Msg("received commit")
+
+		if s.lastTXLSN != 0 && s.lastTXLSN != m.CommitLSN {
+			return fmt.Errorf("out of order commit %s, expected %s", m.CommitLSN, s.lastTXLSN)
+		}
+	default:
+		if xld.WALStart > 0 {
+			s.walWritten = xld.WALStart
```
lovromazgon, line 242: Doesn't this part potentially mess up the check on line 352? In case when `s.Handler` in line 217 doesn't create a record, we will have a LSN in `s.walWritten` that will never match `s.walFlushed`.

If `s.Handler` in line 217 would return a boolean, telling us if it created a record or not, we could decide if this LSN should be written into `s.walWritten` or into `s.lastTXLSN`. In other words - we would be tracking the LSN of any message that requires an ack in `s.walWritten`, while we would be tracking the largest LSN we have received of any other message in `s.lastTXLSN`. If we do that, the naming doesn't fit that well anymore though.

source/logrepl/internal/subscription.go
```diff
@@ -214,9 +218,31 @@ func (s *Subscription) handleXLogData(ctx context.Context, copyDataMsg *pgproto3
...
+			Msg("received begin msg")
+	case *pglogrepl.CommitMessage:
+		sdk.Logger(ctx).Debug().
+			Stringer("commit_lsn", m.CommitLSN).
+			Stringer("tx_lsn", m.TransactionEndLSN).
+			Stringer("expected_end_lsn", s.lastTXLSN).
+			Msg("received commit")
+
+		if s.lastTXLSN != 0 && s.lastTXLSN != m.CommitLSN {
+			return fmt.Errorf("out of order commit %s, expected %s", m.CommitLSN, s.lastTXLSN)
```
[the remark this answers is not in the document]
lyuboxa: The original idea to keep track of the last LSN in a tx, so we can build a good accounting for completion.
I'm not worried about a bug in Postgres but potentially bug in our code or the go code. I see it as a gate.

source/logrepl/internal/subscription.go
lyuboxa: > Doesn't this part potentially mess up the check on line 352? In case when s.Handler in line 217 doesn't create a record, we will have a LSN in s.walWritten that will never match s.walFlushed.

Only if we are not covering all the messages, but for proto v1 we (mostly ) are. The handler will deal with the changes and relation messages and here we deal with the begin/commit to track the tx end LSN to ensure we receive it. 

> If s.Handler in line 217 would return a boolean, telling us if it created a record or not, we could decide if this LSN should be written into s.walWritten or into s.lastTXLSN. In other words - we would be tracking the LSN of any message that requires an ack in s.walWritten, while we would be tracking the largest LSN we have received of any other message in s.lastTXLSN. If we do that, the naming doesn't fit that well anymore though.

Handler returning a signal when the message has been written makes sense, but let's not confuse the meaning of what `walWritten` is compared to begin message end of tx lsn. 

This can change such that the handler can return an error when it hasn't encountered the commit of the stream its processing. Essentially pushing the lastTXLSN in the handler code.

Will takes makes sense:
1. Change Handler to return bool or LSN which described what was written
2. Track begin/commit inside the handler struct

For context

https://www.postgresql.org/docs/current/protocol-replication.html#PROTOCOL-REPLICATION-START-REPLICATION

https://www.postgresql.org/docs/current/protocol-logicalrep-message-formats.html

source/logrepl/internal/subscription.go
```diff
@@ -184,11 +184,15 @@ func (s *Subscription) handlePrimaryKeepaliveMessage(ctx context.Context, copyDa
 	if err != nil {
 		return fmt.Errorf("failed to parse primary keepalive message: %w", err)
 	}
+
+	atomic.StoreUint64((*uint64)(&s.serverWALEnd), uint64(pkm.ServerWALEnd))
```
lovromazgon, line 188: Seems like we can get away without the `atomic` function, since `serverWALEnd` will only ever be read/written by a single goroutine. Right?

lyuboxa: @lovromazgon yea, except my test was causing a data race by reading it for comparison.

source/logrepl/internal/subscription.go
lyuboxa: @lovromazgon addressed here 55f4d7b7eebb98d9143f62c2fbb4767f39983f5b

source/logrepl/internal/subscription.go
```diff
@@ -315,20 +320,33 @@ func (s *Subscription) sendStandbyStatusUpdate(ctx context.Context) error {
...
-		Str("walApply", walFlushed.String()).
+		Stringer("wal_write", s.walWritten).
+		Stringer("wal_flush", walFlushed).
+		Stringer("server_wal_end", s.serverWALEnd).
+		Bool("server_wal_end_sent", replyWithWALEnd).
 		Msg("sending standby status update")
 
-	err := pglogrepl.SendStandbyStatusUpdate(ctx, s.conn, pglogrepl.StandbyStatusUpdate{
+	if replyWithWALEnd {
+		if err := pglogrepl.SendStandbyStatusUpdate(ctx, s.conn, pglogrepl.StandbyStatusUpdate{
```
lovromazgon, line 335: There's an oopsie - if `replyWithWALEnd` is `true` we will send two status updates.

Rather do something like this:
```suggestion
	var statusUpdate pglogrepl.StandbyStatusUpdate
	if replyWithWALEnd {
		statusUpdate = pglogrepl.StandbyStatusUpdate{...}
	} else {
		statusUpdate = pglogrepl.StandbyStatusUpdate{...}
	}
	pglogrepl.SendStandbyStatusUpdate(ctx, s.conn, statusUpdate)
```

lyuboxa: Oh yes fixing.

lyuboxa: Fixed in 900b921e8b6ee400e970389aa11b107b9d26d8ca

source/logrepl/internal/subscription.go
```diff
@@ -315,20 +334,40 @@ func (s *Subscription) sendStandbyStatusUpdate(ctx context.Context) error {
...
+			WALWritePosition: s.serverWALEnd,
+		}); err != nil {
+			return fmt.Errorf("failed to send standby status update with server end lsn: %w", err)
+		}
+
+		return nil
+	}
+
+	// No messages have been flushed, in this case set the flushed
```
lyuboxa, line 358: @lovromazgon This part is kind of tricky, since the last flushed wal on start will be zero since no acks have happened, in which case, we can only safely assume all data before the consistent point or position was flushed.

really ideally, we should be reading the slot and taking the last flushed lsn in there, but this is only applicable to cases where a) the position was entirely lost b) there is an existent slot.

source/logrepl/internal/subscription.go
```diff
@@ -87,14 +91,25 @@ func CreateSubscription(
 			Msgf("replication slot %q already exists", slotName)
 	}
 
+	// If a ConsistentPoint is available use it, otherwise fall back to the startLSN
+	slotLSN, err = pglogrepl.ParseLSN(result.ConsistentPoint)
+	if err != nil {
+		if startLSN == 0 {
+			return nil, fmt.Errorf("failed to parse consistent wal lsn: %w", err)
+		}
```
lovromazgon, line 99: Wait, won't this prevent me to attach a new pipeline to an existing replication slot? 😕 What if we query [`pg_replication_slots.restart_lsn`](https://www.postgresql.org/docs/9.4/catalog-pg-replication-slots.html) and use that as our start LSN (in case no position was provided and we didn't create the slot)?

lyuboxa: Yes. I wanted to do this work as a follow up, because I’d like to refactor what type of connection the subscription is using. Rather than using raw pgconn, it will be best to use pgx so we can take advantage of the querier interface. 

The idea is to not release new version until this is addressed, which will close another issue with invalid LSN in position.

lovromazgon (approved):

PR #171, opened 2024-06-14
Add support for reading replication slot

### LINKED ISSUE #146: Bug: position can be out of range of replication slot
### Bug description

Replication slots can be reused or preconfigured. An issue can occur where the replication slot was replaced during runtime, or the replication slot was preconfigured and modified during connector downtown. In this situation a position can contain an LSN which points to WAL which was cleaned up. 

The solution is to ensure that the position LSN is within the replication slot range during open, when the replication slot exists.

### Steps to reproduce

1. Create table t1
2. Start connector in CDC mode for table t1 and replication slot name slot_t1
3. Insert changes in t1
4. Stop connector
5. Delete replication slot and re-create it.
6. Insert few more changes in t1
7. Resume connector

### Version

latest

lyuboxa:
### Description

* Read the state of the replication slot when subscription is created.
* Ensure the starting position in CDC mode is consistent with what is available in the slot.
* Refactor connection management to use pgxpool and allow for acquired connections to be upgraded for replication
* Add subscription teardown method which stops and waits for wind down.

Fixes #146 

### Quick checks:

### DIFF
--- a/source.go
+++ b/source.go
@@ -22,8 +22,8 @@ import (
 
 	"github.com/conduitio/conduit-commons/csync"
 	"github.com/conduitio/conduit-connector-postgres/source"
+	"github.com/conduitio/conduit-connector-postgres/source/cpool"
 	"github.com/conduitio/conduit-connector-postgres/source/logrepl"
-	cpool "github.com/conduitio/conduit-connector-postgres/source/pool"
 	sdk "github.com/conduitio/conduit-connector-sdk"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
--- a/source.go
+++ b/source.go
@@ -23,6 +23,7 @@ import (
 	"github.com/conduitio/conduit-commons/csync"
 	"github.com/conduitio/conduit-connector-postgres/source"
 	"github.com/conduitio/conduit-connector-postgres/source/logrepl"
+	cpool "github.com/conduitio/conduit-connector-postgres/source/pool"
 	sdk "github.com/conduitio/conduit-connector-sdk"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
@@ -63,7 +64,7 @@ func (s *Source) Configure(_ context.Context, cfg map[string]string) error {
 }
 
 func (s *Source) Open(ctx context.Context, pos sdk.Position) error {
-	pool, err := pgxpool.New(ctx, s.config.URL)
+	pool, err := cpool.New(ctx, s.config.URL)
 	if err != nil {
 		return fmt.Errorf("failed to create a connection pool to database: %w", err)
 	}
--- a/source/logrepl/internal/replication_slot.go
+++ b/source/logrepl/internal/replication_slot.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 Meroxa, Inc.
+// Copyright © 2024 Meroxa, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/source/logrepl/internal/replication_slot.go
+++ b/source/logrepl/internal/replication_slot.go
@@ -1,0 +1,48 @@
+// Copyright © 2022 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import (
+	"context"
+	"errors"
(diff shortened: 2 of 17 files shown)


source.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/conduitio/conduit-commons/csync"
 	"github.com/conduitio/conduit-connector-postgres/source"
 	"github.com/conduitio/conduit-connector-postgres/source/logrepl"
+	cpool "github.com/conduitio/conduit-connector-postgres/source/pool"
```
lovromazgon, line 26: If we always import this as `cpool` we might as well rename the package itself to `cpool` 🤷

source/pool/pool.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+var WithReplCtxKey = struct{}{}
```
lovromazgon, line 25: I'd make this a bit safer and easier to use by hiding the key and exposing a method instead.

```suggestion
type replicationCtxKey struct{}

func WithReplication(ctx context.Context) context.Context {
	return context.WithValue(ctx, replicationCtxKey{}, true)
}
```

source/pool/pool.go
```diff
@@ -0,0 +1,68 @@
+// Copyright © 2022 Meroxa, Inc.
```
lovromazgon, line 1: Not that it matters, just noticed it...
```suggestion
// Copyright © 2024 Meroxa, Inc.
```

source/logrepl/internal/replication_slot.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+
+	"github.com/jackc/pglogrepl"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+var ErrMissingSlot = errors.New("replication slot missing")
+
+type ReplicationSlotResult struct {
```
lovromazgon, line 29: Naming nitpick
```suggestion
type ReadReplicationSlotResult struct {
```

source/logrepl/internal/replication_slot.go
```diff
@@ -0,0 +1,48 @@
+// Copyright © 2022 Meroxa, Inc.
```
lovromazgon, line 1:
```suggestion
// Copyright © 2024 Meroxa, Inc.
```

source/logrepl/internal/replication_slot_test.go
```diff
@@ -0,0 +1,62 @@
+// Copyright © 2022 Meroxa, Inc.
```
lovromazgon, line 1:
```suggestion
// Copyright © 2024 Meroxa, Inc.
```

source/logrepl/internal/subscription.go
```diff
@@ -61,18 +61,27 @@ type Handler func(context.Context, pglogrepl.Message, pglogrepl.LSN) (pglogrepl.
...
 	h Handler,
 ) (*Subscription, error) {
-	var slotLSN pglogrepl.LSN
+	// Request a replication connection
+	conn, err := pool.Acquire(context.WithValue(ctx, cpool.WithReplCtxKey, true))
+	if err != nil {
+		return nil, fmt.Errorf("could not establish replication connection: %w", err)
+	}
+	defer func() { // release connection on error
+		if err != nil {
```
lovromazgon, line 77: To make this work reliably, we need to make the returned error a named variable `err`.

lovromazgon (commented):
I think this makes sense 👍

source.go
lyuboxa: yea, will do!

source/logrepl/internal/replication_slot.go
```diff
@@ -0,0 +1,48 @@
...
+var ErrMissingSlot = errors.New("replication slot missing")
+
+type ReplicationSlotResult struct {
+	Name              string
+	ConfirmedFlushLSN pglogrepl.LSN
+	RestartLSN        pglogrepl.LSN
+}
+
+// ReadReplicationSlot returns state of an existing replication slot.
+func ReadReplicationSlot(ctx context.Context, conn *pgxpool.Pool, name string) (ReplicationSlotResult, error) {
```
lyuboxa, line 36:
```suggestion
func ReadReplicationSlot(ctx context.Context, conn *pgxpool.Pool, name string) (ReadReplicationSlotResult, error) {
```

source/logrepl/internal/subscription.go
lyuboxa: Ooh, I had it and it dropped off in the refactor, thanks for catching that.

source.go
lyuboxa: Fixed in 6708b6d

source/logrepl/internal/replication_slot.go
lyuboxa: Fixed in 6708b6d

source/logrepl/internal/replication_slot_test.go
lyuboxa: Fixed in 6708b6d

source/logrepl/internal/subscription.go
lyuboxa: Fixed in 6708b6d

lovromazgon (approved):
🚀

PR #174, opened 2024-06-22
avro schema generator

lyuboxa:
### Description

This PR adds the functionality to generate an Avro schema for each `snapshot`, `create` and `update` operations. 
The avro schema is rendered in JSON and added to each record under the `postgres.avro.schema` metadata key.

The schema may be used to hint about types where precision or other may be lost as they are converted to go types, e.g. dates, numeric, etc.

Usage of the schema is a temporary measure until Conduit schema management is released.

In order for the schema to be emitted, the following config var will need to be set `logrepl.withAvroSchema`

### Quick checks:


source/config.go
```diff
@@ -74,6 +75,10 @@ type Config struct {
 	// LogreplAutoCleanup determines if the replication slot and publication should be
 	// removed when the connector is deleted.
 	LogreplAutoCleanup bool `json:"logrepl.autoCleanup" default:"true"`
+
+	// WithAvroSchema determines whether the connector should attach an avro schema on each
+	// record.
+	WithAvroSchema bool `json:"logrepl.withAvroSchema" default:"false"`
```
hariso, line 81: Why logrepl only?

source/schema/avro.go
```diff
@@ -0,0 +1,161 @@
...
+		"float8":  avro.NewPrimitiveSchema(avro.Double, nil),
+		"int8":    avro.NewPrimitiveSchema(avro.Long, nil),
+		"int4":    avro.NewPrimitiveSchema(avro.Int, nil),
+		"int2":    avro.NewPrimitiveSchema(avro.Int, nil),
+		"text":    avro.NewPrimitiveSchema(avro.String, nil),
+		"varchar": avro.NewPrimitiveSchema(avro.String, nil),
+		"timestamptz": avro.NewPrimitiveSchema(
+			avro.Long,
+			avro.NewPrimitiveLogicalSchema(avro.TimestampMicros),
+		),
```
hariso, line 48: It doesn't look like Avro has any time type that supports time zones?

source/schema/avro.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (a *avroExtractor) extractType(t *pgtype.Type, val any) (avro.Schema, error) {
+	if ps, ok := a.avroMap[t.Name]; ok {
+		return ps, nil
+	}
+
+	switch tt := val.(type) {
+	case pgtype.Numeric:
+		// N.B.: Default to 38 positions and pick the exponent as the scale.
```
hariso, line 133: Any specific reason for 38?

source/schema/avro_test.go
```diff
@@ -0,0 +1,248 @@
...
+		return cmp.Compare(a.Name(), b.Name())
+	})
+
+	s, err := avro.NewRecordSchema(table, avroNS, fields)
+	is.NoErr(err)
+
+	return s
+}
+
+func avrolizeMap(fields []pgconn.FieldDescription, values []any) map[string]any {
```
hariso, line 223: I love the name.:)

source/schema/avro.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+const (
+	avroNS = "conduit.postgres"
+)
+
+var Avro = &avroExtractor{
+	pgMap: pgtype.NewMap(),
+	avroMap: map[string]*avro.PrimitiveSchema{
```
hariso, line 35: It looks like, at least for now, we don't care if a column is not nullable?

source/schema/avro.go
```diff
@@ -0,0 +1,161 @@
...
+		fs, err := avro.NewFixedSchema(string(avro.Decimal), "",
+			38,
+			avro.NewDecimalLogicalSchema(38, int(math.Abs(float64(tt.Exp)))),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create avro.FixedSchema: %w", err)
+		}
+		return fs, nil
+	default:
+		return nil, fmt.Errorf("cannot resolve field %q of type %T", t.Name, tt)
```
hariso, line 143: I believe that we're intentionally trying to focus only on certain types in this PR, but I just wanted to confirm that we're fine with this list of types.

hariso (commented):
@lyuboxa @samirketema I left a few comments, I'm mostly interested in nullability of columns and that we're fine with the list of supported types for now. Awesome work!

source/config.go
lovromazgon: Do we want to add this to the readme or is this for now just an "internal" feature we use for a special use case?

lyuboxa: > Why logrepl only?

because it is only enabled when in logrepl mode?

source/schema/avro.go
lyuboxa: Yes, all these are right now "required" I guess. Nullability will depend on the pg data type, which I have not added. Is this what you mean?

source/schema/avro.go
lyuboxa: > It doesn't look like Avro has any time type that supports time zones?

No, not at least as a primitive. It will stored as int64 and augmented with a logical primitive.

source/schema/avro.go
lyuboxa: @hariso In this case it is the default scale in Snowflake. In practice because numeric types are limited by the pgx driver to  int64/float64 (at 8 bytes), which really can be accomplished with this being set to 19 (rather than the double 38) 😅 

Now, when a fixed schema is defined with `38` as size, this is the size in bytes. The avro library will use big.Rat as way to encode/decode decimals, so the size will need to be matched to the rational encoded in bytes. It's great.. 38 in this case is absolutely random and it should be ok for now. The way to do this is to set it to the max possible size of big.Rat (which can be excessive) lol

source/schema/avro.go
lyuboxa: For now yea, this default here to remind us harshly if we skip anything.

source/schema/avro.go
hariso: Yup that's what I meant.

source/config.go
hariso: I thought we might want to make this "more generic", but given that we're targeting a very specific use case, it's probably not relevant.:)

source/schema/avro.go
hariso: Can we then use a custom schema to avoid losing the time zone information?

source/schema/avro.go
hariso: Thanks for the explanation! It might be good to just paste this into a comment in the code, as a refresher for future us.:)

source/schema/avro.go
lyuboxa: I think it will be easier to assume this is the time is in UTC all the time and use that to calculate timezone offset? Wdyt?

source/schema/avro.go
lyuboxa: @hariso yea will do.

source/schema/avro.go
lyuboxa: @hariso here is the spec https://avro.apache.org/docs/1.11.1/specification/#timestamp-microsecond-precision it assumes time is in UTC. I think what we need to make sure is that all times are in UTC when passed around in the connector. Same for dates.

lyuboxa: This is already the case as per this type formatter. https://github.com/ConduitIO/conduit-connector-postgres/blob/main/source/types/time.go#L28-L30

hariso: Sounds good!

hariso (approved):

source/schema/avro.go
lyuboxa: see 4df7f87

hariso (approved):

lovromazgon (approved):

PR #175, opened 2024-06-25
UUID as pk

lyuboxa:
### Description

Please include a summary of the change and what type of change it is (new feature, bug fix, refactoring, documentation).
Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

### Quick checks:


source/types/uuid.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+type UUIDFormatter struct{}
+
+// Format returns:
+// * string format of Time when connectorn is not builtin
+// * time type in UTC when connector is builtin
```
hariso, line 27: Comment needs to be updated.

source/types/uuid.go
```diff
@@ -0,0 +1,42 @@
...
+func (UUIDFormatter) Format(v any) (string, error) {
+	b, ok := v.([16]byte)
```
hariso, line 29: So a UUID can come in either as a byte slice (with type set to pgtype.UUIDOID) or as a string (what's the type in that case?)?

PR #176, opened 2024-06-27
don't check relAvroSchema on deletes

samirketema:
### Description

This fixes a nil pointer reference for deletes. Note that [we invoke `updateAvroSchema()` on inserts and updates](https://github.com/ConduitIO/conduit-connector-postgres/blob/5ecd48eadec2f4881a41909942e48fa3149f350d/source/logrepl/handler.go#L142), but not deletes. Therefore, we should not check `relAvroSchema` on delete records.

### Quick checks:


anna-cross (approved):

lyuboxa:
This is fixed in #177 with few more other things

PR #177, opened 2024-06-27
Include before data in payload for updates and deletes; Fix avro rel schema in deletes

lyuboxa:
### Description

When `REPLICA IDENTITY` is set to `FULL`, old values are available to be included in the payload for both `DELETE` and `UPDATE` operations and should be included.

Ensure nil is handled for UUID formatting.

### Quick checks:


raulb (approved):
LGTM

PR #214, opened 2024-10-11
Use bytes instead of [n]bytes to represent decimal type

lyuboxa:
### Description

Since each value may have larger size, as per spec https://avro.apache.org/docs/1.10.2/spec.pdf

```json
{
  "type": "bytes",
  "logicalType": "decimal",
  "precision": 4,
  "scale": 2
}
```

> A decimal logical type annotates Avro bytes or fixed types. The byte array must contain the two's-complement representation of the unscaled integer value in big-endian byte order. The scale is fixed, and is specified using an attribute.

### Quick checks:

### DIFF
--- a/source/schema/avro.go
+++ b/source/schema/avro.go
@@ -25,10 +25,6 @@ import (
 	"github.com/jackc/pgx/v5/pgtype"
 )
 
-const (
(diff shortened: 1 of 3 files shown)


hariso (approved):
:rocket:

PR #220, opened 2024-10-25
add connector logs to tests

alarbada:
### Description

Added logs to the tests. My reason is that I want to see the behavior of the PostgreSQL snapshot so I can mimic it in my MySQL connector. I'd like to see the logs of the operation, and there's no better way than to just run the tests.

### Quick checks:

### DIFF
--- a/test/helper.go
+++ b/test/helper.go
@@ -253,7 +253,12 @@ func IsPgError(is *is.I, err error, wantCode string) {
 }
 
 func Context(t *testing.T) context.Context {
+	ctx := context.Background()
+	if testing.Short() || !testing.Verbose() {
+		return ctx
+	}
+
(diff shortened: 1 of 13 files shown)


test/helper.go
```diff
@@ -250,3 +251,9 @@ func IsPgError(is *is.I, err error, wantCode string) {
 	is.True(ok) // expected err to be a *pgconn.PgError
 	is.Equal(pgerr.Code, wantCode)
 }
+
+func Context(t *testing.T) context.Context {
```
lyuboxa, line 255: One comment, perhaps check if `testing.Verbose()` or `test.Short()` to determine if verbose output is needed. 
Particularly do not enable logging when short output is requested and enable it when verbose requested.

lyuboxa (approved):

PR #234, opened 2024-12-25
Make it possible to disable schemas

hariso:
### Quick checks:


source/snapshot/iterator.go
```diff
@@ -122,8 +123,10 @@ func (i *Iterator) buildRecord(d FetchData) opencdc.Record {
 	metadata["postgres.table"] = d.Table
 
 	rec := sdk.Util.Source.NewRecordSnapshot(pos, metadata, d.Key, d.Payload)
-	cschema.AttachKeySchemaToRecord(rec, d.KeySchema)
-	cschema.AttachPayloadSchemaToRecord(rec, d.PayloadSchema)
+	if i.conf.WithAvroSchema {
```
lyuboxa, line 126: This needs to happen in the cdc handler as well, i.e. don't attach the schema if it is disabled.

hariso: Good callout :+1: Done in 642fb0e. It was pretty much working, because the CDC handling wasn't extracting any schemas. However, we still attempted to attach the schema, which resulted in empty schema names and zero schema versions.

PR #235, opened 2024-12-27
Fix error on empty snapshot

hariso:
### Description

This PR fixes the following bug: when the connector is configured to take a snapshot, and the source table is empty, the source connector fails with the following error:

`can't scan into dest[0]: cannot scan NULL into *int64`.

### Quick checks:


lovromazgon (dismissed):

source/snapshot/fetch_worker.go
```diff
@@ -257,11 +257,10 @@ func (f *FetchWorker) updateSnapshotEnd(ctx context.Context, tx pgx.Tx) error {
 		return nil
 	}
 
-	if err := tx.QueryRow(
-		ctx,
-		fmt.Sprintf("SELECT max(%s) FROM %s", f.conf.Key, f.conf.Table),
-	).Scan(&f.snapshotEnd); err != nil {
-		return fmt.Errorf("failed to query max on %q.%q: %w", f.conf.Table, f.conf.Key, err)
+	query := fmt.Sprintf("SELECT COALESCE(max(%s), 0) FROM %s", f.conf.Key, f.conf.Table)
+	err := tx.QueryRow(ctx, query).Scan(&f.snapshotEnd)
```
lyuboxa, line 261: nit, why not inline the err in the if, to not have to deal with scope

lyuboxa (approved):

source/snapshot/fetch_worker.go
hariso: Makes sense, will do.

PR #247, opened 2025-02-04
Add middleware validation, remove table parameter

hariso:
### Description

- Middleware validation needs to be explicitly called when custom validation is used.
- `table` has been deprecated for some time, so this PR is removing it.

### Quick checks:


maha-hajja (approved):
lgtm! we should let the platform team know about this, just in case they're using the "table" parameter.

PR #248, opened 2025-02-04
change yaml configurations into a table

maha-hajja:
### Description

tables are easier to read :D

### Quick checks:


hariso:
I agree that a table is more readable when looking for information about the parameters (at least for me). The YAML version is better for just copy-pasting the configuration and modifying it.

maha-hajja:
keeping the yaml, but the params order is fixed with the latest sdk

PR #251, opened 2025-02-08
column name && table name are enclosed in quotes to ensure proper handling of capitalized identifiers

derElektrobesen:
### Description

Unfortunately squirrel doesn't support special identifiers quotting: https://github.com/Masterminds/squirrel/issues/276
But in some cases (identifiers with capitalized letters for example) it is required.

This MR adds the support of this kind identifiers.

I'm understand that this kind of names are not "postgress-like", but in some cases it is required.
For example, in out task we need to migrate legacy databases from MySQL (where capitalized identifiers are usual) into Postgress without any renaming.
### Quick checks:


destination.go
```diff
@@ -298,7 +298,7 @@ func (d *Destination) formatUpsertQuery(
 	colArgs, valArgs := d.formatColumnsAndValues(key, payload)
 
 	return d.stmtBuilder.
-		Insert(tableName).
+		Insert(`"` + tableName + `"`).
```
maha-hajja, line 301: @derElektrobesen thanks for the contribution! I have two notes here:
1. let's have a separate method to add the quotes, we have many uses for it. [not picky about this one, but could help with handling special cases, if any]
2. we need to ensure this works for any name, what if the table name has a quote `"` in the middle, or a `'` ? plz double-check for special characters and add some tests for those cases.

derElektrobesen: Fixed
Check again please

derElektrobesen:
@maha-hajja , friendly remind =)

maha-hajja (approved):
@derElektrobesen lgtm! just need to fix the `golangci-lint` and we're good to go

maha-hajja:
@derElektrobesen just change the `golangci-lint` in the `go.mod` file back to this for now:
> 	github.com/golangci/golangci-lint v1.64.4
 and try running `make lint` after.

derElektrobesen:
Locally both v1.64.4 && v1.64.5 golangci-lint versions doesn't show any warnings.
Tried version downgrading to check pipeline status.

hariso:
Looking at the failed job, the error is:

> Error: Failed to run: Error: Command failed: /home/runner/golangci-lint-1.64.4-linux-amd64/golangci-lint config verify

This comes from the GitHub action, which is verifying the config first, and then running the linter (we only run the linter locally).

PR #276, opened 2025-04-25
feat: Implement ReadN

raulb:
### Description

Since the introduction of [end-to-end batching on the Connector SDK](https://github.com/ConduitIO/conduit-connector-sdk/pull/193), (part of the [Connector SDK 0.12.0](https://github.com/ConduitIO/conduit-connector-sdk/releases/tag/v0.12.0)), source connectors can add support to batching by implementing the `ReadN` method. This wasn't a breaking change since the SDK will fallback to the traditional [`Read` method](https://conduit.io/docs/developing/connectors/developing-source-connectors/#read) as it can be spotted in the logs when starting Conduit.

Example:

```bash
2025-04-24T19:21:21+00:00 INF source does not support batch reads, falling back to single reads component=plugin connector_id=postgres-to-kafka:postgres-source plugin_name=builtin:postgres plugin_type=source
```

This pull-request implements `ReadN` while still maintaining support to `Read` method for those who wish to use this connector in an older version of Conduit. This is probably not necessary considering Postgres is a [builtin connector](https://conduit.io/docs/core-concepts/#built-in-connector).

To really utilize the advantage of this, it is recommended to run Conduit with the flag that enables a new upcoming architecture via `--preview.pipeline-arch-v2`. More information on [this blog post](https://meroxa.com/blog/optimizing-conduit-5x-the-throughput/).

#### Running benchmarks

Inserting 20M records using [Benchi](https://github.com/conduitio/benchi) using CDC:

- WIth `Read`: 142824.47 msg/s
- With `ReadN`: 153107.18 msg/s (previously ) 

This is a 7,2% better on CDC.

### Quick checks:


source/snapshot/iterator.go
```diff
@@ -100,6 +100,54 @@ func (i *Iterator) Next(ctx context.Context) (opencdc.Record, error) {
...
 }
 
+// NextN takes and returns up to n records from the queue. NextN is allowed to
+// block until either at least one record is available or the context gets canceled.
+func (i *Iterator) NextN(ctx context.Context, n int) ([]opencdc.Record, error) {
+	if n <= 0 {
+		return nil, fmt.Errorf("n must be greater than 0, got %d", n)
+	}
+
+	var records []opencdc.Record
```
lovromazgon, line 110: There's an optimization we can do here by pre-allocating the slice, otherwise we're creating more allocations than we should, especially for big batches.

Now, what size should we pick? An easy solution would be to just allocate `n` elements and expect that we'll fill the slice up. Of course that will waste space if we don't actually fill the slice up. So I'm thinking that we could store the last batch size in the iterator and use that to allocate the slice, if it's smaller than `n`. The assumption here is that we will mostly see batches of the same size. For instance, if records are steadily streaming and batches are filled, we'll constantly create slices of the full batch. Then, if the stream dies down and we don't have a lot to process, the batch size will decrease and we'll allocate smaller slices.

```suggestion
	records := make([]opencdc.Record, min(i.lastBatchSize, n))
	...
	// before returning the records store the batch size
	i.lastBatchSize = len(records)
	return records, nil
```

Or do you see a better strategy?

source/logrepl/cdc.go
```diff
@@ -116,32 +116,78 @@ func (i *CDCIterator) StartSubscriber(ctx context.Context) error {
...
+			// subscription is done because the context is canceled, we went
+			// into the wrong case by chance
+			return nil, err
+		}
+		// subscription stopped without an error and the context is still
+		// open, this is a strange case, shouldn't actually happen
+		return nil, fmt.Errorf("subscription stopped, no more data to fetch (this smells like a bug)")
+	case rec := <-i.records:
+		recs = append(recs, rec)
+	}
```
hariso, line 165: I tried out a couple of things with channels in [this gist](https://gist.github.com/hariso/e6b5275e534c7ad9c91373203ea2014a) that might help here too. The test sends a number of objects from one goroutine to another through a channel.

The sending goroutine sends 20M objects in batches of different sizes (1-100k). The channel that's shared by the sending and receiving goroutine is either buffered or unbuffered.

On my own machine, sending those 20M objects in batches of 1 record with an unbuffered channel takes around 5.5 seconds.

With batch size set to 1, and a buffered channel of size 50, that drops to 1.8 seconds (3x less).

The biggest improvement though was I set the batch size to 1k or 10k (it didn't matter much if the channel was buffered), and the result was around 80ms.

With that, I think it's worth if we try sending the data from the CDC and snapshot iterator in batches.

hariso: This is what I meant with iterators sending data in batches: https://github.com/ConduitIO/conduit-connector-postgres/compare/read-n...haris/read-n-batches?expand=1

source/logrepl/combined.go
```diff
@@ -134,6 +135,77 @@ func (c *CombinedIterator) Next(ctx context.Context) (opencdc.Record, error) {
...
+
+				// Retry with new iterator
+				return c.NextN(ctx, n)
+			}
+			return nil, fmt.Errorf("failed to fetch records in batch: %w", err)
+		}
+		return records, nil
+	}
+
+	// Fall back to using Next() if NextN is not implemented
```
lovromazgon, line 167: I'd simply remove `Next` and not make our lives harder than they are. This code is internal to the connector, so we're not breaking anything.

Also, since `CombinedIterator` just forwards the call to either the `CDCIterator` or the `SnapshotIterator`, I don't think we need the code to fetch the remaining (n-1) records, the actual iterator will already do that for us.

source/logrepl/cdc_test.go
```diff
@@ -485,6 +485,132 @@ func TestCDCIterator_Ack(t *testing.T) {
...
+			is.NoErr(err)
+		}
+
+		// This requests more than available
+		records, err := i.NextN(ctx, 5)
+		is.NoErr(err)
+
+		// Verify we got fewer records than requested
+		// and that they are in order
+		is.True(len(records) > 0 && len(records) <= 2)
```
lovromazgon, line 547: The assertion made me look twice. So it can happen that the iterator returns either 1 or 2 records, right? I'd change the test to call `NextN` again until we have 2 records. And, additionally, I'd issue _another_ call to `NextN` where we don't expect anything to get back (we can create a context deadline in that case to make sure we don't wait forever). Note that I didn't check if there is maybe already a test that tests what happens if `NextN` doesn't have anything to fetch - if we do, we don't have to do the last call of course.

Along these lines:

```suggestion
		// As long as we don't have 2 records, keep fetching records
		var records []opencdc.Record
		for len(records) < 2 {
			recordsTmp, err := i.NextN(ctx, 5)
			is.NoErr(err)
			records = append(records, recordsTmp...)
		}
		ctxWithTimeout, cancel := context.WithTimeout(ctx, 500*time.Millisecond)
		defer cancel()
		_, err = i.NextN(ctxWithTimeout, 5)
		is.True(errors.Is(err, context.DeadlineExceeded))
```

source/logrepl/cdc.go
lovromazgon: Good point, we should try collecting records as soon as we can to cut down on the code that processes records 1 by 1 as much as we can.

source/snapshot/iterator.go
hariso: >  Of course that will waste space if we don't actually fill the slice up. 

How much space would actually be wasted? 

There's another thing that plays a role and that is the `snapshot.fetchSize` parameter (which I think probably can be removed, because it we have `sdk.batch.size`). While the iterator is active, we can always expect the database to return that amount of records (except for the last fetch),

raulb: @hariso good point. I have actually been confused by the two whenever I used this connector as a source. This change would require deprecating the field, making use only of the SDK parameter, but I'll see how to approach this in a way that doesn't increase much scope.

source/logrepl/cdc_test.go
lyuboxa: This is sensible

source/logrepl/combined.go
```diff
@@ -111,27 +111,29 @@ func NewCombinedIterator(ctx context.Context, pool *pgxpool.Pool, conf Config) (
...
+	records, err := c.activeIterator.NextN(ctx, n)
 	if err != nil {
 		// Snapshot iterator is done, handover to CDC iterator
-		if !errors.Is(err, snapshot.ErrIteratorDone) {
-			return opencdc.Record{}, fmt.Errorf("failed to fetch next record: %w", err)
-		}
-
-		if err := c.useCDCIterator(ctx); err != nil {
-			return opencdc.Record{}, err
+		if errors.Is(err, snapshot.ErrIteratorDone) {
```
lyuboxa, line 125: Nitpick, but why increase the nesting here?

source/snapshot/iterator.go
lyuboxa: This is a bit more nuanced than the CDC iterator, because records are coming in the fetch size defined in the config. As it stands, each record from that cursor fetch is fed 1 by 1 into the channel. If there is caching at that level, you essentially need to take N and not have to pre-allocate.

source/logrepl/combined.go
raulb: @lyuboxa i had actually changed how these errors operated by switching to CDC explicitly calling to `.NextN(` [later on](https://github.com/ConduitIO/conduit-connector-postgres/pull/276/files?diff=unified&w=0#diff-16bc8373489d2c228c26cf08f43674185eacb1d97804697ef714742a5335238aR132). I realize that maybe it's better actually switch to CDC when the snapshot iterator finishes, so I'll change this back a bit.

source/snapshot/iterator.go
lyuboxa: > There's another thing that plays a role and that is the `snapshot.fetchSize` parameter (which I think probably can be removed, because it we have `sdk.batch.size`). While the iterator is active, we can always expect the database to return that amount of records (except for the last fetch),

@hariso I'm unclear about this, the fetchSize is the cursor fetch size, e.g. how many rows to get from Postgres, this is primarily used to optimize retrieval of rows. Removing that will result in `select * from table` which may or may not perform better, but something we should try, but somewhat tangential to this PR.

source/logrepl/combined.go
raulb: @lyuboxa https://github.com/ConduitIO/conduit-connector-postgres/pull/276/commits/89f2f0caeb188f4cc6c3d73e307ff75eb2eb7d59

source/logrepl/cdc.go
lyuboxa: > Good point, we should try collecting records as soon as we can to cut down on the code that processes records 1 by 1 as much as we can.

This may work ok without a timer, since I think "collecting" needs two parameters (size, *t*<sub>interval</sub>) for which the requirements are satisfied. Now, there is a side-effect which can be used and it has to do with Postgres sending keep alive messages so often that the time may not be required and we can count time based on the keep alive messages.. something like:

````
1. I want 3 messages
2. recv message -> record (n++) 
3. n < 3 
4. recv meessage -> keepalive (t < waitFor) -> continue
5. recv message -> record (n++)
6. n < 3
7. recv message -> keep alive (t >= waitFor) -> return 2 messages
```

This is mostly implementation detail

source/snapshot/iterator.go
lovromazgon: > How much space would actually be wasted?

If you have batches of 10k records, but fill them up with 1 record, you're wasting 99.99% of the space you're allocating 🙃 Using `unsafe.Sizeof(opencdc.Record{})` the empty record is determined to be 104 bytes, so allocating 9999 of them is just shy of 1MB. In the worst case, when records trickle in just slow enough, this can happen tens or hundreds of times per second, putting quite a lot of pressure on the garbage collector.

hariso: @lyuboxa what I meant is to keep the `FETCH`, but not have a `fetchSize` parameter in the connector. Instead, we should use `FETCH sdk.batch.size`.

lyuboxa: @hariso I think this will be a bit confusing, since `sdk.batch.size` is a general config, while fetch size is specific for the snapshot.

hariso: Why does it need to be specific? When I find `sdk.batch.size: 100` in a source connector's configuration, I expect the connector to read 100 records at a time, which is the fetch size in snapshots, or no?

lyuboxa: Because it does not apply to CDC. Its specific to postgres and it may become irrelevant if we decided to move to `select * from table` ¯\_(ツ)_/¯

lyuboxa: @hariso and I had a chat, the TL;DR is that `sdk.batch.size` is what conduit passes to `ReadN` and that may change on configuration, thus allocating a slice with `n` capacity based on the `ReadN` argument makes sense.

```go
records := make([]opencdc.Record, 0, sdkBatchSize)
```

The compiler has no idea how to allocate this on the stack, so it will go on the heap anyway. 

The append to that should be fairly efficient and you can either return the full slice or something like `records[:nread]` but that doesn't matter, the slice will have the same capacity. You need to get that from the source batch configuration from the config.

What @lovromazgon is suggesting will be needed if we are returning `[]opencdc.Record` but I don't think we need it in this implementation.

raulb: Based on my understanding, the first size `n` should always be based on the batch size configured by the SDK anyways (`ReadN`) so the allocation would still be `n`. i.e.: I agree with the last comment by @lyuboxa . Unless I'm missing something else here.

source/logrepl/cdc.go
raulb: I created this issue to track this separately https://github.com/ConduitIO/conduit-connector-postgres/issues/278.

source/logrepl/cdc.go
```diff
@@ -113,35 +113,63 @@ func (i *CDCIterator) StartSubscriber(ctx context.Context) error {
...
-// block until either a record is returned from the subscription, the
-// subscription stops because of an error or the context gets canceled.
-// Returns error when the subscription has been started.
-func (i *CDCIterator) Next(ctx context.Context) (opencdc.Record, error) {
+// NextN takes and returns up to n records from the queue. NextN is allowed to
+// block until either at least one record is available or the context gets canceled.
+func (i *CDCIterator) NextN(ctx context.Context, n int) ([]opencdc.Record, error) {
 	if !i.subscriberReady() {
-		return opencdc.Record{}, errors.New("logical replication has not been started")
+		return []opencdc.Record{}, errors.New("logical replication has not been started")
```
lovromazgon, line 120: Now that we're returning a slice, we can return `nil` if we're returning an error.

```suggestion
		return nil, errors.New("logical replication has not been started")
```

Same for the other lines that return an error, also in the snapshot and combined iterators.

source/logrepl/cdc.go
```diff
@@ -113,35 +113,63 @@ func (i *CDCIterator) StartSubscriber(ctx context.Context) error {
...
 	}
 
-	for {
+	if n <= 0 {
+		return []opencdc.Record{}, fmt.Errorf("n must be greater than 0, got %d", n)
+	}
+
+	var recs []opencdc.Record
```
lovromazgon, line 127: Let's pre-allocate the slice here as well.

source.go
```diff
@@ -119,8 +119,12 @@ func (s *Source) Open(ctx context.Context, pos opencdc.Position) error {
 	return nil
 }
 
-func (s *Source) Read(ctx context.Context) (opencdc.Record, error) {
-	return s.iterator.Next(ctx)
+func (s *Source) ReadN(ctx context.Context, n int) ([]opencdc.Record, error) {
+	if n <= 0 {
+		return []opencdc.Record{}, nil
+	}
```
lovromazgon, line 125: I know you're trying to be careful, but I'd just remove this to prevent other developers from thinking it needs to be checked. The SDK should provide the correct `n` that is always at least `1`, otherwise that should be considered a bug in the SDK.

lovromazgon (commented):
A few cleanups, otherwise looks good 👍

source/snapshot/iterator.go
lovromazgon: Allocating size `n` is fine with me, better than not allocating! 👍

source.go
raulb: @lovromazgon yeah, you're right. I recall seeing that check in the SDK.

source/logrepl/cdc.go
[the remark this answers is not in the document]
raulb: @hariso yup, changing all of them now.

source/logrepl/cdc.go
```diff
@@ -113,35 +113,63 @@ func (i *CDCIterator) StartSubscriber(ctx context.Context) error {
...
+// block until either at least one record is available or the context gets canceled.
+func (i *CDCIterator) NextN(ctx context.Context, n int) ([]opencdc.Record, error) {
 	if !i.subscriberReady() {
-		return opencdc.Record{}, errors.New("logical replication has not been started")
+		return []opencdc.Record{}, errors.New("logical replication has not been started")
 	}
 
-	for {
+	if n <= 0 {
+		return []opencdc.Record{}, fmt.Errorf("n must be greater than 0, got %d", n)
```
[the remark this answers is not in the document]
raulb: Yup

source/logrepl/cdc.go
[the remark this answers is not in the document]
hariso: PS: didn't notice Lovro's comments about the same thing in another place. Sorry for the spam.:D

source/logrepl/cdc_test.go
```diff
@@ -485,6 +456,138 @@ func TestCDCIterator_Ack(t *testing.T) {
...
+			is.NoErr(err)
+		}
+
+		var allRecords []opencdc.Record
+
+		// Collect records until we have all 3
+		for len(allRecords) < 3 {
+			attemptCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			records, err := i.NextN(attemptCtx, 3-len(allRecords))
+			cancel()
```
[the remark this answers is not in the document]
raulb: @hariso actually, we can't since the we're canceling the context right after we collect records, right?

source/logrepl/cdc.go
```diff
@@ -113,35 +113,63 @@ func (i *CDCIterator) StartSubscriber(ctx context.Context) error {
...
+	}
+
+	for len(recs) < n {
 		select {
+		case rec := <-i.records:
+			recs = append(recs, rec)
 		case <-ctx.Done():
-			return opencdc.Record{}, ctx.Err()
+			// Return what we have with the error
+			return recs, ctx.Err()
```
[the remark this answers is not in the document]
raulb: @hariso maybe not. I looked at the previous method and thought we had returned the collected record, but now that I look at it, I see we were returning an empty record, so it's better to return nil here as well.

source/logrepl/cdc_test.go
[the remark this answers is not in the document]
hariso: So, let me try understanding the purpose first, and that is: we want to make sure we don't wait for the data for too long, right? Do we need cancel the context at all?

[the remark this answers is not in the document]
raulb: @hariso touché https://github.com/ConduitIO/conduit-connector-postgres/pull/276/commits/5f9fdc2909cfc1d82294d18b0eacd4fbb9366495

source/logrepl/cdc_test.go
```diff
@@ -485,6 +456,138 @@ func TestCDCIterator_Ack(t *testing.T) {
...
+		for j := 1; j <= 2; j++ {
+			_, err := pool.Exec(ctx, fmt.Sprintf(`INSERT INTO %s (id, column1, column2, column3, column4, column5)
+				VALUES (%d, 'test-%d', %d, false, 12.3, 14)`, table, j+20, j, j*100))
+			is.NoErr(err)
+		}
+
+		// Will keep calling NextN until all records are received
+		var records []opencdc.Record
+		for len(records) < 2 {
+			recordsTmp, err := i.NextN(ctx, 5)
```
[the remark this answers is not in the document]
raulb: https://github.com/ConduitIO/conduit-connector-postgres/pull/276/commits/2385595918abc4ba2567bf02a6db9f74b8fbcfa4

hariso (approved):

PR #284, opened 2025-05-06
Optimize ReadN in CDC by collecting records in batches

### LINKED ISSUE #278: Feature: Add support for collecting records in batches
### Feature description

At the moment, we're collecting records 1 by 1 when we should sending them in batches to optimize the use of the channel used to send records between go routines. 

[Here's a thread](https://github.com/ConduitIO/conduit-connector-postgres/pull/276/files/5aec436a79c8e25ed31696ca91b728bc72f1465a#r2066123548) for more context and an implementation by @hariso [here](https://github.com/ConduitIO/conduit-connector-postgres/compare/read-n...haris/read-n-batches?expand=1).

hariso:
### Description

Closes #278.

### Quick checks:


source/logrepl/cdc_test.go
```diff
@@ -575,17 +662,12 @@ func TestCDCIterator_NextN(t *testing.T) {
...
-		}()
+		time.Sleep(100 * time.Millisecond)
+		is.NoErr(i.Teardown(ctx))
 
-		records, err := i.NextN(ctx, 5)
-		if err != nil {
-			is.True(strings.Contains(err.Error(), "logical replication error"))
-		} else {
-			is.True(len(records) > 0)
-		}
```
hariso, line 588: When I read this part of the test, I wasn't sure what the expected behavior was (i.e., whether we expected an error or not). Based on the test description ("subscription termination"), it appeared to me that an error is what we should always expect.

source/logrepl/handler.go
```diff
@@ -33,22 +35,75 @@ import (
...
+
+func (h *CDCHandler) scheduleFlushing(ctx context.Context) {
+	ticker := time.NewTicker(h.flushInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
```
hariso, line 81: This isn't quite correct, since it will flush the records no matter when the last flush happened (which can be less that `flushInterval` ago, if the batch was flushed because it reached `batchSize`). I'll change this.

source/logrepl/cdc.go
```diff
@@ -124,9 +153,45 @@ func (i *CDCIterator) NextN(ctx context.Context, n int) ([]opencdc.Record, error
 		return nil, fmt.Errorf("n must be greater than 0, got %d", n)
 	}
 
-	var recs []opencdc.Record
+	// First, we check if there are any records from the previous batch
+	// that we can start with.
+	recs := make([]opencdc.Record, len(i.recordsForNextRead), n)
```
raulb, line 158: It'd be interesting to see if this PR implementation fixes https://github.com/ConduitIO/conduit-connector-postgres/issues/282.

source/logrepl/internal/subscription.go
```diff
@@ -193,7 +194,7 @@ func (s *Subscription) listen(ctx context.Context) error {
 
 		copyDataMsg, ok := msg.(*pgproto3.CopyData)
 		if !ok {
-			return fmt.Errorf("unexpected message type %T", msg)
+			return fmt.Errorf("unexpected message type %T, value: %v", msg, msg)
```
raulb, line 197: Good call.

source/logrepl/cdc.go
```diff
@@ -142,33 +207,59 @@ func (i *CDCIterator) NextN(ctx context.Context, n int) ([]opencdc.Record, error
...
+
+		return batch, nil
+	default:
+		// No more records currently available
+		return nil, nil
 	}
+}
 
-	return recs, nil
+// appendRecordsWithLimit moves records from src to dst, until the given limit is reached,
```
raulb, line 245:
```suggestion
// appendRecordsWithLimit appends records to dst from src, until the given limit is reached,
```

source/logrepl/cdc.go
```diff
@@ -64,8 +78,22 @@ func NewCDCIterator(ctx context.Context, pool *pgxpool.Pool, c CDCConfig) (*CDCI
...
+	// The handler can prepare the next batch, and the CDCIterator can use them
+	// to return the maximum number of records.
+	batchesCh := make(chan []opencdc.Record, 1)
+	handler := NewCDCHandler(
+		ctx,
+		internal.NewRelationSet(),
+		c.TableKeys,
+		batchesCh,
+		c.WithAvroSchema,
+		c.BatchSize,
```
raulb, line 93: I'm wondering if we could have a default value for this one in case it's not set as part of the configuration in `CDCConfig`. @hariso What do you think?

i.e.: This wouldn't be necessary https://github.com/ConduitIO/conduit-connector-postgres/pull/284/files.

raulb (approved):
@hariso Left some comments, but approving because I don't consider them blockers. I'm particularly interested in the benchmark results comparing to our latest tests due to:

1. Implementation is supposedly more performant actually collecting records in batches.
2. See if the change has any impact on https://github.com/ConduitIO/conduit-connector-postgres/issues/282

source/logrepl/cdc.go
hariso: @raulb which file specifically are you referring to in the link?

hariso:
> @hariso Left some comments, but approving because I don't consider them blockers. I'm particularly interested in the benchmark results comparing to our latest tests due to:
> 
>     1. Implementation is supposedly more performant actually collecting records in batches.
> 
>     2. See if the change has any impact on [Investigate: Why allocating memory is less efficient on `ReadN` #282](https://github.com/ConduitIO/conduit-connector-postgres/issues/282)

I've just run in on my EC2 instance (c7i.xlarge), and the CDC test that we currently have in `streaming-benchmarks` gives me cca. 64k msg/s, whereas this version of the connector gives around cca. 74k msg/s. I ran both versions twice.

source/logrepl/cdc.go
raulb: @hariso LOL, I grabbed the wrong link, sorry about that 🤦‍♂️. I meant this https://github.com/ConduitIO/conduit-connector-postgres/issues/282

raulb: @hariso I don't seem to find the line I was referring to. Never mind 🤷‍♂️

PR #290, opened 2025-05-19
Supporting reading and writing numeric/decimal values without loss of precision

hariso:
### Description

Closes https://github.com/ConduitIO/conduit-connector-postgres/issues/292.

This PR adds support for reading `numeric`/`decimal` values and writing those. The values are read into `big.Rat` values. `big.Rat` values are also written into `numeric`/`decimal` columns in the database.

To convert a `big.Rat` value (which is a fractional number) into a numeric value, we need to calculate the ratio up to a given precision, which is retrieved from the database. The precision/scale from the database is cached and expires after 5 minutes.

### Quick checks:

### DIFF
--- a/destination.go
+++ b/destination.go
@@ -379,15 +379,15 @@ func (d *Destination) formatValue(table string, column string, val interface{}) 
 
 // formatBigRat formats a big.Rat into a string that can be written into a NUMERIC/DECIMAL column.
 func (d *Destination) formatBigRat(table string, column string, v *big.Rat) (string, error) {
+	if v == nil {
+		return "", nil
+	}
+	
 	// we need to get the scale of the column so we that we can properly
 	// round the result of dividing the input big.Rat's numerator and denominator.
 	scale, err := d.dbInfo.GetNumericColumnScale(table, column)
 	if err != nil {
 		return "", fmt.Errorf("failed getting scale of numeric column: %w", err)
-	}
-
-	if v == nil {
-		return "", nil
 	}
 
 	//nolint:gosec // no risk of overflow, because the scale in Pg is always <= 16383
--- a/destination.go
+++ b/destination.go
@@ -219,7 +219,7 @@ func (d *Destination) insert(ctx context.Context, r opencdc.Record, b *pgx.Batch
 		return err
 	}
 
-	colArgs, valArgs, err := d.formatColumnsAndValues(tableName, key, payload)
+	colArgs, valArgs, err := d.formatColumnsAndValues(ctx, tableName, key, payload)
 	if err != nil {
 		return fmt.Errorf("error formatting columns and values: %w", err)
 	}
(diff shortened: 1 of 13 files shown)


source/types/numeric.go
```diff
@@ -15,42 +15,23 @@
...
-		}
-
-		v, err := i8v.Value()
-		if err != nil {
-			return nil, err
-		}
-
-		return v, nil
+// Format converts a pgtype.Numeric to a big.Rat.
+func (NumericFormatter) Format(num pgtype.Numeric) (*big.Rat, error) {
```
raulb, line 26: Maybe we could rename this method to make it more explicit. e.g.: `BigRatFromNumeric` (or something like that)

internal/db_info.go
```diff
@@ -0,0 +1,112 @@
...
+
+// tableCache stores information about a table.
+// The information is cached and refreshed every 'cacheExpiration'.
+type tableCache struct {
+	columns     map[string]int
+	nextRefresh time.Time
+}
+
+// cacheExpiration defines how long table information remains valid
+const cacheExpiration = 5 * time.Minute
```
raulb, line 40: Could we make this configurable instead? To let the user control this.

destination.go
```diff
@@ -346,3 +365,31 @@ func (d *Destination) getKeyColumnName(key opencdc.StructuredData, defaultKeyNam
...
+	// we need to get the scale of the column so we that we can properly
+	// round the result of dividing the input big.Rat's numerator and denominator.
+	scale, err := d.dbInfo.GetNumericColumnScale(table, column)
+	if err != nil {
+		return "", fmt.Errorf("failed getting scale of numeric column: %w", err)
+	}
+
+	if v == nil {
+		return "", nil
+	}
```
raulb, line 391: Should this be done before instead maybe?

internal/db_info.go
```diff
@@ -0,0 +1,112 @@
...
+		return 0, fmt.Errorf("error querying column info: %w", err)
+	}
+
+	// Check if the column is of the numeric/decimal type
+	if dataType != "numeric" && dataType != "decimal" {
+		return 0, fmt.Errorf("column %s in table %s is not a numeric type (actual type: %s)", column, table, dataType)
+	}
+
+	// Handle case where numeric_scale is NULL
+	if numericScale == nil {
```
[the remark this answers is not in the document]
hariso: Not really, because `numericScale` is what we get from the database:

```go
var numericScale *int

	// todo use proper ctx
	err := d.conn.QueryRow(context.Background(), query, table, column).Scan(&dataType, &numericScale)
```

i.e. the numeric scale from the database is read into `numericScale`.

internal/db_info.go
hariso: Hmm, that's an option too. However, I don't know if a user would find this useful or what would drive their decision. Schema changes are very infrequent, and you can't really predict them, which means that configuring them is also likely to be just guessing.

lyuboxa: I don't think this needs to be exposed to the user.

lyuboxa (approved):

PR #296, opened 2025-06-06
Composite key support in destination

lovromazgon:
### Description

This PR adds support for writing records to tables with composite keys. Note that this removes the `key` setting in the destination connector. I don't see why we need it, since we require the key to contain structured data, which is a map, and the key in the map represents the column name (the same as in `.Payload.After`).

Also I noticed a flaky test in the logrepl package where a closed channel was closed. Turns out that closing of that channel is not even necessary, no reader relies on that channel being closed, so I just removed that to make it safer.

### Quick checks:

### DIFF
--- a/destination.go
+++ b/destination.go
@@ -76,7 +76,7 @@ func (d *Destination) Write(ctx context.Context, recs []opencdc.Record) (int, er
 	b := &pgx.Batch{}
 	for _, rec := range recs {
 		var err error
-		rec, err = d.cleanRecord(rec)
+		rec, err = d.ensureStructuredData(rec)
 		if err != nil {
 			return 0, fmt.Errorf("failed to clean record: %w", err)
 		}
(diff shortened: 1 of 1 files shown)


destination.go
```diff
@@ -343,22 +281,51 @@ func (d *Destination) formatColumnsAndValues(ctx context.Context, table string,
...
+	if !ok {
+		return false
 	}
-	for k := range key {
-		return k
+	return len(structuredKey) > 0
+}
+
+// cleanRecord makes sure the record key and payload are structured data.
+func (d *Destination) cleanRecord(r opencdc.Record) (opencdc.Record, error) {
```
raulb, line 293: May I suggest maybe naming this method `ensureStructuredRecord`?

destination.go
```diff
@@ -74,6 +76,10 @@ func (d *Destination) Write(ctx context.Context, recs []opencdc.Record) (int, er
 	b := &pgx.Batch{}
 	for _, rec := range recs {
 		var err error
+		rec, err = d.cleanRecord(rec)
```
raulb, line 79: Left a suggestion about renaming this method below (name it `ensureStructuredRecord`)

raulb (approved):
Only left a naming suggestion. Approving in the meantime anyways.

destination.go
lovromazgon: You may 🙇

